PR #21, opened 2019-03-08
Drop travis ci.

thomasvl:
Drop travis ci.

bazel ci on buildkite is providing everything now.


allevato:
Looks like you have the same problem I ran into on rules_swift; you need to go into the GitHub branch protection settings and disable the checks that don't have `/pr` in them, because changing the Buildkite settings causes those to never be sent anymore.

PR #51, opened 2019-10-11
Remove CI tests for versioned releases of Bazel and dependencies.

swiple-rules-gardener:
Remove CI tests for versioned releases of Bazel and dependencies.

Unfortunately, the team maintaining this repository does not have the resources to keep the versioned releases up to date. Going forward, the recommended approach for depending on this repository is to select a specific commit and to use the `git_repository` WORKSPACE rule available in `@bazel_tools//tools/build_defs/repo:git.bzl`.


README.md
```diff
@@ -12,10 +12,13 @@ Add the following to your `WORKSPACE` file to add this repository as a dependenc
 ```python
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
+# apple_support no longer supports releases. If you'd like to pin down these
+# dependencies to a specific commit, please add the following to the top of your
+# WORKSPACE, using the commit you'd like to pin the repository.
```
justhecuke: There appears to be tests for`macos_last_green`. Can you add instructions on how to find the commit hash for that task?

sergiocampama: can you elaborate? what do you mean by "commit hash for that task"?

justhecuke: I think I misunderstood what that task was doing. With this change, we can be relatively sure that every commit in apple_support "works" with the last green build of bazel?

sergiocampama: no, it just makes sure that `apple_support` works with which ever version of Bazel happened to be the green build at the time. it may be that some time after, that same commit won't work on a newer version of the last green build, but hopefully by that time master has been updated to work with that version of bazel

justhecuke: OK. So there's a mapping of `apple_support` commit to `bazel` commit where the tests passed. How do we get at that mapping to find a commit that works for the commit of bazel that we are using?

sergiocampama: ah, I see what you mean. can you see the BuildKite checks here? https://buildkite.com/bazel/apple-support-darwin/builds/330

If so, open the last task and then open the Bazel info section. That should point to the commit:
Screen Shot 2019-10-11 at 6 21 23 PM

PR #63, opened 2020-08-11
Use stardoc to generate documentation

### LINKED ISSUE #62: Add docs generation
https://github.com/bazelbuild/stardoc

segiddins:
Closes #62 

Not sure if this is a desirable change, though, because it noticeably changes the output format -- also, I don't think startdoc yet has a way to generate the README.md from a given set of `stardoc` targets


thomasvl:
- For the README.md, you'll want to manually create it.
- You can use [header_template](https://github.com/bazelbuild/stardoc/blob/master/docs/stardoc_rule.md#stardoc-header_template) to provide a real header on the page, but it does mean you have to provide a template per page.  Looking at the startdoc source, it does support pulling the module comment in, but the default template doesn't use it.
- You also will run into https://github.com/bazelbuild/stardoc/issues/58, the java can parse this out of the doc strings, but it isn't exposed to the template.
- You're also likely to run into https://github.com/bazelbuild/stardoc/pull/60

ps - you can use a sh_test (or a golden_test rule) to check the generated files vs. the checked in one if you wanted.

rules/BUILD
```diff
@@ -2,6 +2,20 @@ load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 
 licenses(["notice"])
 
+exports_files([
+    "apple_genrule.bzl",
+])
+
+bzl_library(
```
thomasvl, line 9: what was this needed for?

doc/BUILD.bazel
```diff
@@ -0,0 +1,29 @@
+load("@io_bazel_stardoc//stardoc:stardoc.bzl", "stardoc")
+
+stardoc(
```
thomasvl, line 3: It you put these in the directories with the bzl files, you won't have to export the bzl files.

https://github.com/bazelbuild/stardoc/issues/57 is opened for the rule to maybe better support bzl_library directly for srcs.

PR #67, opened 2020-10-30
Remove passing lists of args to genrule

keith:
This is required for this migration https://github.com/bazelbuild/bazel/issues/5903

### DIFF
--- a/rules/apple_genrule.bzl
+++ b/rules/apple_genrule.bzl
@@ -80,7 +80,7 @@ def _apple_genrule_impl(ctx):
         inputs = resolved_srcs.to_list() + resolved_inputs,
         outputs = files_to_build.to_list(),
         env = ctx.configuration.default_shell_env,
-        command = argv,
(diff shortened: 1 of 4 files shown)


rules/apple_genrule.bzl
```diff
@@ -80,7 +80,7 @@ def _apple_genrule_impl(ctx):
         inputs = resolved_srcs.to_list() + resolved_inputs,
         outputs = files_to_build.to_list(),
         env = ctx.configuration.default_shell_env,
-        command = argv,
+        command = " ".join(argv),
```
segiddins, line 83: should we use the shell escape method in skylib here? https://github.com/bazelbuild/bazel-skylib/blob/182046f090d6ecae264c7e5d520da096370287e2/lib/shell.bzl#L37-L49

keith: looks like that overquotes

brentleyjones (approved):

PR #68, opened 2020-10-31
Rename build files BUILD.bazel

HadrienPatte:
See https://github.com/bazelbuild/bazel/issues/4517


keith:
Thanks for submitting this, but unless there's a bit more general consensus in the bazel community about this, specifically for our case we're better off matching whatever google is doing since they push commits to the upstream branch that we optionally cherry pick, so renaming this would make that process harder

PR #90, opened 2021-05-10
Add multi_arch_wrapper for macOS CLIs

keith:
This is a rule + transition for forcing macOS CLIs are built with
multiple architectures, even in the host configuration (with `cfg =
"exec"`). This is useful to share caches across Intel and Apple Silicon
machines, and to be able to populate the cache from either machine as
well. This needs to live in apple_support instead of rules_apple so that
rules_swift can use it for the Swift worker.

### DIFF
--- a/rules/multi_arch_wrapper.bzl
+++ b/rules/multi_arch_wrapper.bzl
@@ -1,0 +1,65 @@
+"""
+Wrap a macOS tool in a multi-arch transition. This is useful for building fat
+binaries to share caches between Apple Silicon and Intel machines. This
+behavior requires Xcode 12
+"""
+
+# NOTE: This lives in apple_support instead of rules_apple so that rules_swift
+# can depend on it
+
+def _force_multi_arch_transition_impl(settings, attr):
+    return {"//command_line_option:macos_cpus": "arm64,x86_64"}
+
+_force_multi_arch_transition = transition(
+    implementation = _force_multi_arch_transition_impl,
+    inputs = [],
+    outputs = ["//command_line_option:macos_cpus"],
+)
+
+def _multi_arch_wrapper_impl(ctx):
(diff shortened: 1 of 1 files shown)


keith:
up for a different name here

rules/multi_arch_wrapper.bzl
```diff
@@ -0,0 +1,65 @@
...
+Wrap a macOS tool in a multi-arch transition. This is useful for building fat
+binaries to share caches between Apple Silicon and Intel machines. This
+behavior requires Xcode 12
+"""
+
+# NOTE: This lives in apple_support instead of rules_apple so that rules_swift
+# can depend on it
+
+def _force_multi_arch_transition_impl(settings, attr):
+    return {"//command_line_option:macos_cpus": "arm64,x86_64"}
```
thii, line 11: Tried this but it doesn't work for host tools for me. Seems like host compilation doesn't read the `--macos_cpus` flag..

keith: With this transition it doesn't respect it at all and just hard codes this. Are you saying that you're still not getting fat binaries if it's host?

thii: I'm not getting fat binaries if it's host (only tried the transition, not the rule)

keith: I'll have to test again, I thought this worked for host tools but maybe it did or something regressed

thii: It works if I use the rule directly.

thii:
Submitted https://github.com/bazelbuild/apple_support/pull/95 as an alternative.

keith:
https://github.com/bazelbuild/apple_support/pull/95

PR #95, opened 2021-11-11
Add `universal_binary` rule to build multi-arch macOS binaries

thii:
This is basically a copy of the new `apple_universal_binary` rule but
with a simpler interface (there are no `minimum_os_version` and
`platform_type` attributes), and the output binary is forced to be
multi-arch. This rule is meant to be used for host/exec tools so that
they are built as universal binaries regardless the architecture of the
execution platform.

### DIFF
--- a/test/universal_binary_test.bzl
+++ b/test/universal_binary_test.bzl
@@ -1,0 +1,114 @@
+# Copyright 2021 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Definition of a test rule to test universal_binary."""
+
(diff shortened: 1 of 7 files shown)


brentleyjones (approved):

keith:
One thing I liked about the transition I had before was it should be cross platform. Maybe we can loosen this one to allow that too?

keith:
Theoretically it could just ignore this in some cases, at least for now

thii:
By cross-platform you mean macOS and Linux? or other Apple platforms?

keith:
Linux ,just so you can use the same top level rule

test/universal_binary_test.bzl
```diff
@@ -0,0 +1,114 @@
...
+SYMBOLS=(
+  {binary_contains_symbols}
+)
+
+if [[ ! -f "{file}" ]]; then
+  echo "{file} doesn't exist"
+  exit 1
+fi
+
+actual_symbols=$(objdump -t -macho -arch=x86_64 -arch arm64 {file} | grep -v "*UND*" | awk '{{print substr($0,index($0,$5))}}')
```
keith, line 35: You might be able to do this same thing with `nm -Uj`

test/universal_binary_test.bzl
```diff
@@ -0,0 +1,114 @@
...
+# limitations under the License.
+
+"""Definition of a test rule to test universal_binary."""
+
+# Template for the test script used to validate that the action outputs contain
+# the expected values.
+_TEST_SCRIPT_CONTENTS = """\
+#!/bin/bash
+
+set -eu
```
keith, line 22:
```suggestion
set -euo pipefail
```

keith (approved):

PR #104, opened 2022-01-22
Add documentation generation

keith:
This currently only generates the rules.md file, but it's a start! This
also updates the manually edited READMEs with new links to these.

This is a bit weird because the apple_genrule that folks use directly is
a macro, so we expose the docs for the underlying rule instead.

This was based on what we have in rules_swift and rules_apple

Fixes https://github.com/bazelbuild/apple_support/issues/62


rules/rules.doc.bzl
```diff
@@ -0,0 +1,9 @@
+"""
+Helper rules for Apple platforms
+"""
+
+load("//rules/private:apple_genrule.bzl", _apple_genrule = "apple_genrule")
+load("//rules:universal_binary.bzl", _universal_binary = "universal_binary")
+
+apple_genrule = _apple_genrule
+universal_binary = _universal_binary
```
keith, line 9: this file isn't meant to be included but needs to exist to centralize the rules for stardoc so we have a single markdown file for both rules

thii (approved):

PR #112, opened 2022-03-17
Minor doc typo

erikkerber:


keith:
Thanks! can you update the original too? https://github.com/bazelbuild/apple_support/blob/2ec0ef3eb2954371596fbb1a225fb244fcde31a9/lib/apple_support.bzl#L329 we should be generating this one but we're not currently, but it was at one point by google I believe

erikkerber:
🙇🏼 

I'll go ahead and patch my well-earned boycott badge on my shirt

keith (approved):

PR #113, opened 2022-03-23
Add Apple CC toolchain setup

keith:
This migrates the Apple CC toolchain from bazel into this repo. This allows this to evolve and be updated without a full bazel udpate. Bazel's built in unix toolchain can still be used to build macOS C++ targets, but this repo handles a wider variety of Apple specific flags and requires you have Xcode installed.


keith:
Investigation related to https://github.com/bazelbuild/bazel/issues/15041#issuecomment-1076719673

keith:
Overview / approach of changes:

- I copied all the macOS toolchain setup I could from bazel without changing it, so most files like wrapped_clang, xcrunwrapper etc have 0 changes
- I moved all the tests that required non-macOS only support, or relied on actually building objc, without changing them (objc_test.sh, apple_test.sh, the relevant helpers)
- I fixed buildifier violations as needed
- I copied the file that is now setup.bzl but changed it significantly since we don't need to handle the same sort of multi-os fallbacks
- I plan to cleanup some things like unnecessary env vars later so we can have more clear history on why they're not needed etc

crosstool/setup.bzl
```diff
@@ -0,0 +1,109 @@
...
+        "HOMEBREW_RUBY_PATH",
+        "SYSTEMROOT",
+        "USER",
+    ] + MSVC_ENVVARS,
+    implementation = _apple_cc_autoconf_impl,
+    configure = True,
+)
+
+# buildifier: disable=unnamed-macro
+def apple_cc_configure():
```
thii, line 97: I think we can keep the same comment https://github.com/bazelbuild/bazel/blob/4ec0757f00f6aa66a37122f020984c3a6c9f2949/tools/cpp/cc_configure.bzl#L182
```suggestion
def apple_cc_configure():
    """A C++ configuration rules that generate the crosstool file."""
```

README.md
```diff
@@ -1,14 +1,59 @@
...
+supports Bazel 6.
+
+To use the Apple CC toolchain, pull this repository into your build and
+add this to your `.bazelrc`:
+
+```bzl
+build --enable_platform_specific_config
+build:macos --apple_crosstool_top=@local_config_apple_cc//:toolchain
+build:macos --crosstool_top=@local_config_apple_cc//:toolchain
+build:macos --host_crosstool_top=@local_config_apple_cc//:toolchain
```
thii, line 37:
```suggestion
build --apple_crosstool_top=@local_config_apple_cc//:toolchain
build --crosstool_top=@local_config_apple_cc//:toolchain
build --host_crosstool_top=@local_config_apple_cc//:toolchain
```

crosstool/setup.bzl
```diff
@@ -0,0 +1,109 @@
+"""Configure the Apple CC toolchain"""
+
+load("//crosstool:osx_cc_configure.bzl", "configure_osx_toolchain")
+
+_DISABLE_ENV_VAR = "BAZEL_NO_APPLE_CPP_TOOLCHAIN"
```
thii, line 5: If anyone doesn't want to use the autodetected Apple C++ toolchain, they can just not pass all the custom `--*crosstool_top` flags. Is there still a need for this env var?

README.md
keith: If you build on Linux I don't think you should set these there though?

crosstool/setup.bzl
keith: Yea I only left it for debugging, I didn't put it in the docs. We could underscore it if we wanted it to look more private

README.md
thii: Ahh I thought this is meant for the macOS platform. So this is like a base config for all Apple platforms? I think we can just mention the flags instead of an example .bazelrc.

keith: it is, but if your project builds on both platforms, setting this globally might cause issues

crosstool/BUILD.toolchains
```diff
@@ -0,0 +1,100 @@
+package(default_visibility = ["//visibility:public"])
+
+# Target constraints for each arch.
+# TODO(apple-rules): Rename osx constraint to macOS.
```
brentleyjones, line 4: Is there migration needed for this, or what is preventing us from doing this now?

crosstool/setup.bzl
```diff
@@ -0,0 +1,109 @@
...
+    "VS90COMNTOOLS",
+    "VS100COMNTOOLS",
+    "VS110COMNTOOLS",
+    "VS120COMNTOOLS",
+    "VS140COMNTOOLS",
+    "VS150COMNTOOLS",
+    "VS160COMNTOOLS",
+    "TMP",
+    "TEMP",
+]
```
brentleyjones, line 58: Why do we still include these?

brentleyjones (approved):

crosstool/BUILD.toolchains
keith: don't think so anymore since @platforms was forced on folks and they're an alias here, i just copied this one without modification

crosstool/setup.bzl
keith: we don't, i'll do in a follow up since i wanted this to be mostly a copy-paste

googlewalt:
We see a bunch of failures: https://buildkite.com/bazel/google-bazel-presubmit/builds/64142.  We're fixing it in the internal import patch, but I want to check whether this should have been caught by bazel presubmits earlier?

PR #114, opened 2022-04-14
Test CI

keith:


keith:
i was testing pre-commit.ci but I forgot the buildifier hook downloads at hook time so it would depend on https://github.com/pre-commit/pre-commit/issues/1453 to work

PR #128, opened 2022-08-18
Add `//lib:swizzle_absolute_xcttestsourcelocation` target

brentleyjones:
When this target is linked into an XCTest binary, and `BUILD_WORKSPACE_DIRECTORY` is set, the relative paths of `XCTSourceCodeLocation` are made absolute, fixing Test Issue navigation in Xcode.

### DIFF
--- a/lib/BUILD
+++ b/lib/BUILD
@@ -2,6 +2,11 @@ load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 
 licenses(["notice"])
 
+# This target is mainly for use by rules_apple and rules_swift.
+# When this target is linked into an XCTest binary, and the
+# `BUILD_WORKSPACE_DIRECTORY` environment variable is set, the relative paths of
+# `XCTSourceCodeLocation` are made absolute, fixing Test Issue navigation in
+# Xcode.
 objc_library(
     name = "swizzle_absolute_xcttestsourcelocation",
     testonly = True,


lib/BUILD
```diff
@@ -2,6 +2,14 @@ load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
...
 licenses(["notice"])
 
+objc_library(
+    name = "swizzle_absolute_xcttestsourcelocation",
+    testonly = True,
+    srcs = ["swizzle_absolute_xcttestsourcelocation.m"],
+    visibility = ["//visibility:public"],
+    alwayslink = True,
+)
+
```
brentleyjones, line 12: How should we document this?

We need to mention how `BUILD_WORKSPACE_DIRECTORY` needs to be set. Also its main use will be through the `apple.swizzle_absolute_xcttestsourcelocation` `feature` that `rules_apple` and rules_swift` will honor and automatically add this dependency to the test targets.

brentleyjones: (And by how, I also mean literally. I don't think it will get picked up by Stardoc, so at which location as well.)

keith: If this is only going to be referenced internally in the other rules maybe just document in those test rules?

keith (approved):

lib/BUILD
brentleyjones: That's the main use case, so I'm fine with that. Technically it's public API, so I'll add something right above this target for adventurous folks.

PR #137, opened 2022-09-27
Skip `swizzle_absolute_xcttestsourcelocation` on non-macOS platforms

BalestraPatrick:
This target isn't compatible with any other platform other than macOS so we're now declaring its compatibility. This was discovered during the bzlmod integration: https://github.com/bazelbuild/bazel-central-registry/pull/196

### DIFF
--- a/lib/BUILD
+++ b/lib/BUILD
@@ -11,6 +11,12 @@ objc_library(
     name = "swizzle_absolute_xcttestsourcelocation",
     testonly = True,
     srcs = ["swizzle_absolute_xcttestsourcelocation.m"],
+    target_compatible_with = [
+        "@platforms//os:macos",
+        "@platforms//os:watchos",
+        "@platforms//os:ios",
+        "@platforms//os:tvos",
+    ],
     visibility = ["//visibility:public"],
     alwayslink = True,
 )


lib/BUILD
```diff
@@ -13,6 +13,9 @@ objc_library(
     srcs = ["swizzle_absolute_xcttestsourcelocation.m"],
     visibility = ["//visibility:public"],
     alwayslink = True,
+    target_compatible_with = [
+        "@platforms//os:macos",
```
keith, line 17: does this work for iOS still, since the target isn't macos?

brentleyjones: This is similar to what I had to put into rules_swift. Seems like we really need a "is apple platform" check.

BalestraPatrick: Assuming you mean [this](https://github.com/bazelbuild/rules_swift/pull/878/files), how is that different if we were to do this here? What's the preferred approach?
```
target_compatible_with = [
    "@platforms//os:macos",
    "@platforms//os:watchos",
    "@platforms//os:ios",
    "@platforms//os:tvos",
]
```

brentleyjones: Did that way not work?

BalestraPatrick: So I went the same way that rules_swift did in the beginning, but that seems to break on non-macOS platforms (in https://github.com/bazelbuild/bazel-central-registry/pull/196): https://buildkite.com/bazel/bcr-presubmit/builds/541#0183a163-1ec0-4329-b265-8aca36c17d33 

Which makes me wonder if this is working in rules_swift at all. I'll try to go with the `platforms` solution instead now.

brentleyjones: The Linux CI started to pass after we did it that way, so I'm pretty sure it's working there.

BalestraPatrick: Reverted this to use platforms and it seems to work in the central-registry too: https://github.com/bazelbuild/bazel-central-registry/pull/196

BalestraPatrick:
@keith @brentleyjones The patched version was already merged in the central-registry and seems to be working fine, so I'd say we should go ahead with this in this repo and remove the patch in the next release from the central-registry.

brentleyjones (approved):

PR #143, opened 2022-10-14
Add release workflow

keith:
This workflow is manually triggered, takes a tag name, and creates a
release. Before tagging the release it pushes an update to the
MODULE.bazel file. GitHub Actions don't allow multi-line text input, so
it creates the release with TODOs that need to be filled out after it is
created. This still saves quite a few steps in the process.


keith:
example of what this creates on my test repo 
image

thii (approved):

PR #162, opened 2023-01-17
Hardcode host_system_name

keith:
Not sure if this should ever differ, but for now this makes it more
clear that this variable isn't actually used


brentleyjones:
Hmm. Why doesn't this change to arm64 on Apple silicon?

keith:
im not sure what this value is used for so i didn't want to mess with it in this

brentleyjones (approved):

PR #167, opened 2023-01-17
Update allowed env vars

keith:
Most of these variables were windows or unix specific. This leaves just
the ones we actually use with comments on their purpose

### DIFF
--- a/crosstool/setup.bzl
+++ b/crosstool/setup.bzl
@@ -41,54 +41,14 @@ def _apple_cc_autoconf_impl(repository_ctx):
     else:
         repository_ctx.file("BUILD", "# Apple CC autoconfiguration was disabled because you're not on macOS")
 
-MSVC_ENVVARS = [
-    "BAZEL_VC",
-    "BAZEL_VC_FULL_VERSION",
-    "BAZEL_VS",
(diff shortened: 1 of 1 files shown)


crosstool/setup.bzl
```diff
@@ -41,54 +41,14 @@ def _apple_cc_autoconf_impl(repository_ctx):
...
-        "CC_CONFIGURE_DEBUG",
-        "CC_TOOLCHAIN_NAME",
-        "CPLUS_INCLUDE_PATH",
-        "DEVELOPER_DIR",
-        "GCOV",
-        "HOMEBREW_RUBY_PATH",
-        "SYSTEMROOT",
-        "USER",
-    ] + MSVC_ENVVARS,
+        "USE_CLANG_CL",  # Kept as a hack for those who rely on this invaliding the toolchain
```
brentleyjones, line 47: :heart:

brentleyjones (approved):

crosstool/setup.bzl
keith: i think we should just add a new one here called XCODE_VERSION or something, but wasn't going to remove this one until we figured that out

PR #169, opened 2023-01-17
Add new _DYNAMIC_LINK_ACTIONS variable

keith:
`all_link_actions` is actually misleading as it doesn't include
Objective-C links or static links, this replaces this common use case
with a new variable. More work needs to be done to remove
`all_link_actions`

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -1421,7 +1421,7 @@ def _impl(ctx):
         name = "apple_env",
         env_sets = [
             env_set(
-                actions = [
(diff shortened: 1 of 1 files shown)


crosstool/cc_toolchain_config.bzl
```diff
@@ -2582,7 +2569,7 @@ def _impl(ctx):
                 flag_groups = [flag_group(flags = ["-Werror"])],
             ),
             flag_set(
-                actions = all_link_actions,
+                actions = _DYNAMIC_LINK_ACTIONS,
```
keith, line 2572: This is the only one with a behavior change because the bad naming tricked me

brentleyjones (approved):

PR #170, opened 2023-01-17
Use _DYNAMIC_LINK_ACTIONS in more places

keith:


crosstool/cc_toolchain_config.bzl
```diff
@@ -2377,10 +2371,7 @@ def _impl(ctx):
         name = "generate_linkmap",
         flag_sets = [
             flag_set(
-                actions = [
```
keith, line 2380: This one is a change, right now bazel only ever sets this feature if it's objc, but in the future as the objc specific linking infra goes away I think this is the right setup. no downside in the meantime since it'll just never be enabled in that case

brentleyjones (approved):

PR #173, opened 2023-01-17
Add some test binary rules

keith:
This makes it easy to test a few different configurations


keith:
depends on https://github.com/bazelbuild/apple_support/pull/171

thii (approved):

PR #179, opened 2023-01-20
Add config_settings for platforms

keith:
This is a common pattern you need to select on. Since we have a limited
set of CPUs it's nice to provide a source of truth for this type of
thing

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -24,6 +24,7 @@ filegroup(
     testonly = 1,
     srcs = [
         "WORKSPACE",
+        "//configs:for_bazel_tests",
         "//constraints:for_bazel_tests",
         "//crosstool:for_bazel_tests",
         "//lib:for_bazel_tests",
--- a/configs/BUILD
+++ b/configs/BUILD
@@ -1,0 +1,79 @@
+load("@bazel_skylib//lib:selects.bzl", "selects")
+
+package(default_visibility = ["//visibility:public"])
+
+selects.config_setting_group(
+    name = "apple",
+    match_any = [
+        ":darwin",
+        ":darwin_arm64",
+        ":darwin_x86_64",
+        ":ios_arm64",
+        ":ios_arm64e",
+        ":ios_armv7",
+        ":ios_armv7s",
+        ":ios_i386",
+        ":ios_sim_arm64",
+        ":ios_x86_64",
+    ],
+)
+
+# Configuration settings to make doing selects for Apple vs non-Apple platforms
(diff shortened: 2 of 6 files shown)


keith:
looking for feedback, not sure if we should do this or not, but i figured i'd submit for discussion (right now it's missing many cpus)

jszumski:
I've replicated something similar into a couple different projects already, so a single canonical list seems like a great idea 👍

keith:
Can you share what you have for this so we make sure we cover it?

jszumski:
```
[
    config_setting(
        name = type,
        values = {"apple_platform_type": type},
    )
    for type in [
        "ios",
        "macos",
        "tvos",
        "watchos",
    ]
]

[
    config_setting(
        name = cpu,
        values = {"cpu": cpu},
    )
    for cpu in [
        "darwin_x86_64",
        "darwin_arm64",
        "darwin_arm64e",
        "ios_i386",
        "ios_x86_64",
        "ios_armv7",
        "ios_armv7s",
        "ios_arm64",
        "ios_arm64e",
        "ios_sim_arm64",
        "tvos_x86_64",
        "tvos_arm64",
        "tvos_sim_arm64",
        "watchos_arm64_32",
        "watchos_armv7k",
        "watchos_i386",
        "watchos_x86_64",
    ]
]

selects.config_setting_group(
    name = "any_device",
    match_any = [
        ":ios_armv7",
        ":ios_armv7s",
        ":ios_arm64",
        ":ios_arm64e",
        ":tvos_arm64",
        ":watchos_arm64_32",
        ":watchos_armv7k",
    ],
)

selects.config_setting_group(
    name = "any_simulator",
    match_any = [
        ":ios_i386",
        ":ios_x86_64",
        ":ios_sim_arm64",
        ":tvos_x86_64",
        ":tvos_sim_arm64",
        ":watchos_i386",
        ":watchos_x86_64",
        ":watchos_arm64",
    ],
)
```

cgrindel:
I am about to add this one to `swift_bazel`.

```python
[
    config_setting(
        name = type,
        values = {"apple_platform_type": type},
    )
    for type in [
        "ios",
        "macos",
        "tvos",
        "watchos",
    ]
]
```

keith:
ok i added those constraints here except for the apple_platform_type one. i don't feel strongly but i feel like we should leave that out since ideally things moves towards platforms instead and we don't rely on that as much going forward. but happy to add if folks think we should

PR #187, opened 2023-02-13
Add bzl load visibility

keith:
https://bazel.build/versions/6.0.0/concepts/visibility#load-visibility

PiperOrigin-RevId: 506300061
(cherry picked from commit 9b2c2ff6649b6040388a15cdf3fc6ac7d8e71ff1)


keith:
won't be able to take this indefinitely as we support old versions, also doesn't matter that much for us, we can apply to the files as we want in the future instead

PR #190, opened 2023-02-13
Add coverage_prefix_map feature

keith:
This removes absolute paths from code coverage data. This is also
supported in swift with `swift.coverage_prefix_map`


keith:
maybe this should be the default? the problem is that this breaks coverage in xcode at the cost of being non hermetic

brentleyjones:
I think to should be the default.

PR #191, opened 2023-02-15
Add BAZEL_SKIP_XCODE_FETCH to reduce toolchain configuration time

keith:
Since this toolchain requires Xcode, running xcode_locator during
toolchain setup is potentially duplicative. The only things it's used
for is to check that they exist, and to add to the allowed include
directories. The include directories already includes `/Applications`
which is where _most_ people likely have Xcode installed, so as long
as that can be relied on, fetching these values isn't necessary. This
saves a few seconds when the toolchain is reconfigured.

https://bazelbuild.slack.com/archives/CD3QY5C2X/p1652819625121889

### DIFF
--- a/crosstool/osx_cc_configure.bzl
+++ b/crosstool/osx_cc_configure.bzl
@@ -156,9 +156,13 @@ def configure_osx_toolchain(repository_ctx):
         Label("@build_bazel_apple_support//crosstool:wrapped_clang.cc"),
     ))
 
-    (xcode_toolchains, xcodeloc_err) = run_xcode_locator(repository_ctx, xcode_locator)
-    if not xcode_toolchains:
-        return False
+    skip_xcode_fetch = "BAZEL_SKIP_XCODE_FETCH" in repository_ctx.os.environ and repository_ctx.os.environ["BAZEL_SKIP_XCODE_FETCH"] == "1"
+    xcode_toolchains = []
+    xcodeloc_err = ""
+    if not skip_xcode_fetch:
+        (xcode_toolchains, xcodeloc_err) = run_xcode_locator(repository_ctx, xcode_locator)
+        if not xcode_toolchains:
+            return False
 
     # For Xcode toolchains, there's no reason to use anything other than
     # wrapped_clang, so that we still get the Bazel Xcode placeholder


brentleyjones (dismissed):

keith:
maybe instead of this we should flip this around and do an env var if you want to allow Xcode versions outside of `/Applications`. i think that if you had one somewhere else and we didn't do the logic we're doing here to add it to the include paths, bazel would catch it for undeclared header inclusions

keith:
here's what that would look like https://github.com/bazelbuild/apple_support/pull/197

crosstool/osx_cc_configure.bzl
```diff
@@ -156,9 +156,13 @@ def configure_osx_toolchain(repository_ctx):
         Label("@build_bazel_apple_support//crosstool:wrapped_clang.cc"),
     ))
 
-    (xcode_toolchains, xcodeloc_err) = run_xcode_locator(repository_ctx, xcode_locator)
-    if not xcode_toolchains:
-        return False
+    skip_xcode_fetch = "BAZEL_SKIP_XCODE_FETCH" in repository_ctx.os.environ and repository_ctx.os.environ["BAZEL_SKIP_XCODE_FETCH"] == "1"
```
cgrindel, line 159: What do you think about using [bool](https://bazel.build/rules/lib/globals#bool) to evaluate the environment variable value? It allows folks to pick their favorite representation.

cgrindel:
The one upside to this implementation vs #197. It may seem like a more understandable change in behavior for those who run into it. Of course, folks have to opt into the benefit. Perhaps, we can recommend that it be enabled in the [useful flags doc](https://github.com/ios-bazel-users/ios-bazel-users/blob/master/UsefulFlags.md).

crosstool/osx_cc_configure.bzl
keith: this is what the upstream CC toolchain vars enforce so I figure we might as well keep the same behavior

keith:
going with https://github.com/bazelbuild/apple_support/pull/197 in the spirit of trying to have the right defaults

PR #197, opened 2023-02-16
Add BAZEL_ALLOW_NON_APPLICATIONS_XCODE to run xcode-locator

keith:
Since this toolchain requires Xcode, running xcode_locator during
toolchain setup is likely duplicative. The only things it's used
for is to check that they exist, and to add to the allowed include
directories. The include directories already includes /Applications
which is where most people likely have Xcode installed, so as long
as that can be relied on, fetching these values isn't necessary. This
saves a few seconds when the toolchain is reconfigured.

bazelbuild.slack.com/archives/CD3QY5C2X/p1652819625121889

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,10 +77,11 @@ toolchain works. Here are some of the more commonly useful ones:
   changes, which ensures that toolchain binaries will be rebuilt with
   the new version of Xcode so that caches are correctly shared across
   machines.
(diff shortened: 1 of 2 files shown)


keith:
same as https://github.com/bazelbuild/apple_support/pull/191 but with a flipped default, idk which is best, this is less safe but saves some perf

brentleyjones (approved):

brentleyjones:
We should document this somewhere.

PR #198, opened 2023-02-16
Exclude some tests from bcr

keith:
These tests currently fail because of runfiles issues. Really we don't
care that the BCR tests much anyways, just that the file loads work at
all. Our CI covers the rest.

https://bazelbuild.slack.com/archives/C014RARENH0/p1676506451712139


brentleyjones (approved):

keith:
doing this for the failed release too https://github.com/bazelbuild/bazel-central-registry/pull/447

brentleyjones (approved):

PR #204, opened 2023-02-16
Remove pre Xcode 12 logic

keith:
We require Xcode 13+ (matching the App Store) at this point, so we don't
need to support versions that can't build for Apple Silicon

### DIFF
--- a/crosstool/osx_cc_configure.bzl
+++ b/crosstool/osx_cc_configure.bzl
@@ -46,37 +46,6 @@ def _get_escaped_xcode_cxx_inc_directories(repository_ctx, xcode_toolchains):
 
     return include_dirs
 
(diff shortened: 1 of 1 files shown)


crosstool/osx_cc_configure.bzl
```diff
@@ -46,8 +46,7 @@ def _get_escaped_xcode_cxx_inc_directories(repository_ctx, xcode_toolchains):
 
     return include_dirs
 
-# TODO: Remove once Xcode 12 is the minimum supported version
-def _compile_cc_file_single_arch(repository_ctx, src_name, out_name):
```
keith, line 50: diff looks weird but this is deleting this function, and moving the error handling to the one below

brentleyjones (approved):

PR #206, opened 2023-02-17
Flip defaults of hermetic debug info flags

keith:
This flips the defaults of 3 features:

- relative_ast_path: this makes sure that the swift AST entries embedded
  in the final binaries uses relative paths for hermeticity. This could
  potentially break debugging for users relying on absolute paths, in
  which case they should use `platform settings -w "$PWD"` to change the
  lldb PWD so the relative paths are valid, or they should disable this
  feature.
- oso_prefix_is_pwd: this is similar to above except its for N_OSO
  entries in the final binary. The same workarounds apply
- remap_xcode_path: this removes absolute paths to Xcode from the
  binary. This shouldn't affect debugging because lldb's defaults should
  fall back to the same paths even if they aren't in the debug info

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -1657,6 +1657,7 @@ def _impl(ctx):
 
     relative_ast_path_feature = feature(
         name = "relative_ast_path",
+        enabled = True,
         env_sets = [
             env_set(
                 actions = _DYNAMIC_LINK_ACTIONS,
@@ -1777,6 +1778,7 @@ def _impl(ctx):
 
     remap_xcode_path_feature = feature(
         name = "remap_xcode_path",
+        enabled = True,
         flag_sets = [
             flag_set(
                 actions = [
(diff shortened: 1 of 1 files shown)


keith:
feedback wanted, not sure if we should flip or not

mattrobmattrob (approved):
We've used `relative_ast_path` & `oso_prefix_is_pwd` in our build for a while. Seems reasonable to me.

brentleyjones:
> in which case they should use platform settings -w "$PWD" to change the lldb PWD so the relative paths are valid, or they should disable this feature.

Technically it should be the execution root, as `$PWD` only works if the convenience symlinks exist and haven't been renamed: https://github.com/buildbuddy-io/rules_xcodeproj/blob/09ab6abe998e7cf191bb34c8a675438538e27405/xcodeproj/internal/bazel_integration_files/create_lldbinit.sh#L18-L22

brentleyjones (approved):
I agree with these changes: https://github.com/buildbuddy-io/rules_xcodeproj/blob/09ab6abe998e7cf191bb34c8a675438538e27405/xcodeproj/internal/xcodeproj.template.bazelrc#L31-L38

keith:
if anyone spots other issues please report them

PR #212, opened 2023-03-10
[RFC] Add support for alternate linkers

keith:
This is a RFC for adding support for using a different linker than the
default based on a new `--feature`. This is a theoretically replacement
for solutions like [`rules_apple_linker`](https://github.com/keith/rules_apple_linker/).

In some ways this is better, it's easier for users to integrate with,
without a separate rule set, and it's more clear what linker is going to
be used since it's explicitly global vs something in your dependency
tree. Also there is no ambiguity like there is with multiple linkers in
the dependency tree with rules_apple_linker.

This current iteration has some potential UX downsides, like it's harder
to select() on specific linker arguments, so users have to do that
themselves in their macros, but you cannot select() based on features.

Overall I think something like this is worth doing because many users
are interested in at least experimenting with this, and this decreases
the overhead to do so.

### DIFF
--- a/crosstool/BUILD.tpl
+++ b/crosstool/BUILD.tpl
@@ -104,6 +104,13 @@ cc_toolchain_suite(
 %{cxx_builtin_include_directories}
         ],
         tool_paths_overrides = {%{tool_paths_overrides}},
+        alternate_linker_path = %{alternate_linker_path},
+        alternate_linker_args = [
+%{alternate_linker_args}
+        ],
+        default_linker_args = [
+%{default_linker_args}
+        ],
     )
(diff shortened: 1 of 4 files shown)


thii:
See https://github.com/bazelbuild/rules_apple/pull/1925/files

You can also set it globally in your .bazelrc file:

```
build linkopt=--ld-path=external/com_github_keith_ld64.lld/ld64.lld
```

keith:
Yea. I think the UX there is still a bit worse than getting this global setup since you have to make sure to set that on all your macros. Maybe I'm overthinking it though

PR #218, opened 2023-05-04
Remove support for bitcode

keith:
Now that Xcode 14.1 is the minimum supported to the App Store, and Xcode
14.0 dropped support for bitcode, we don't need to handle these flags
anymore.


keith:
https://github.com/bazelbuild/bazel/issues/18321

mattrobmattrob (approved):

PR #219, opened 2023-05-05
Update C++ standard default to 14

keith:
Mirroring https://github.com/bazelbuild/bazel/pull/18280 so it's easier
to swap between the toolchains


keith:
we should only merge if that upstream change does

keith:
hrm maybe we should wait closer to 7.x merging, kinda a lose lose

PR #227, opened 2023-07-11
Add stripped down apple_verification_test

keith:
This is a useful test rule copied from rules_apple stripped down to only
what we need right now. This will allow us to remove some of our shell
tests, and write starlark tests instead of shell tests in the future.


keith:
we can't move the rules_apple one here because it uses providers that are only in rules_aplpe

brentleyjones (approved):

PR #228, opened 2023-07-11
Add initial visionOS support

keith:
Depends on https://github.com/bazelbuild/bazel/pull/18905 and https://github.com/bazelbuild/platforms/pull/71

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -683,9 +683,6 @@ def _impl(ctx):
         ctx.attr.cpu == "ios_i386" or
         ctx.attr.cpu == "ios_sim_arm64" or
         ctx.attr.cpu == "ios_x86_64" or
-        ctx.attr.cpu == "visionos_arm64" or
-        ctx.attr.cpu == "visionos_sim_arm64" or
-        ctx.attr.cpu == "visionos_x86_64" or
         ctx.attr.cpu == "watchos_arm64_32" or
         ctx.attr.cpu == "watchos_armv7k" or
         ctx.attr.cpu == "watchos_i386" or
@@ -725,7 +722,15 @@ def _impl(ctx):
             ],
         )
     else:
-        apply_default_compiler_flags_feature = None
+        apply_default_compiler_flags_feature = feature(
+            name = "apply_default_compiler_flags",
+            flag_sets = [
+                flag_set(
+                    actions = [ACTION_NAMES.objc_compile, ACTION_NAMES.objcpp_compile],
+                    flag_groups = [flag_group(flags = ["-fno-autolink"])],
+                ),
+            ],
+        )
 
     runtime_root_flags_feature = feature(
         name = "runtime_root_flags",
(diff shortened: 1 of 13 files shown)


test/test_data/BUILD
```diff
@@ -34,6 +34,34 @@ universal_binary(
 
 starlark_apple_binary(
     name = "apple_binary",
```
mattrobmattrob, line 36: Perhaps rename this in a follow up to be consistent?
```suggestion
    name = "macos_binary",
```

mattrobmattrob (approved):

crosstool/cc_toolchain_config.bzl
```diff
@@ -675,8 +681,11 @@ def _impl(ctx):
         ctx.attr.cpu == "ios_arm64e" or
         ctx.attr.cpu == "ios_armv7" or
         ctx.attr.cpu == "ios_i386" or
-        ctx.attr.cpu == "ios_x86_64" or
         ctx.attr.cpu == "ios_sim_arm64" or
+        ctx.attr.cpu == "ios_x86_64" or
+        ctx.attr.cpu == "visionos_arm64" or
+        ctx.attr.cpu == "visionos_sim_arm64" or
+        ctx.attr.cpu == "visionos_x86_64" or
```
brentleyjones, line 688: This doesn't seem right? Shouldn't it have it's own for `-DOS_VISIONOS`? (Also, shouldn't watchOS have its own?)

keith: Looks like I took this from the original one https://github.com/bazelbuild/bazel/blob/c3cb69e11629455192650976a7026d51f7eedf3d/tools/osx/crosstool/cc_toolchain_config.bzl#L720-L736

maybe the intent was more "this platform is iOS like"

I think we should just not add visionOS here, I think folks should be using the platform provided ones of these instead anyways, pushed that

brentleyjones:
> Depends on https://github.com/bazelbuild/bazel/pull/18905

Does that mean when we merge this that we drop support for anything before Bazel 7?

crosstool/cc_toolchain_config.bzl
```diff
@@ -716,7 +722,15 @@ def _impl(ctx):
...
-        apply_default_compiler_flags_feature = None
+        apply_default_compiler_flags_feature = feature(
+            name = "apply_default_compiler_flags",
+            flag_sets = [
+                flag_set(
+                    actions = [ACTION_NAMES.objc_compile, ACTION_NAMES.objcpp_compile],
+                    flag_groups = [flag_group(flags = ["-fno-autolink"])],
+                ),
+            ],
+        )
```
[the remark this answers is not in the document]
keith: it's not right now since the else case is actually unreachable but we can force it to be exhaustive instead to make that clear for the future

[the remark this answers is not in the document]
keith: pushed make it exhaustive so we at least have to reconsider that for new CPUs

keith:
Well if you backport that to 6.x maybe we'd only have to worry about 5.x (assuming it's reasonable to only support the newest 6.x). otherwise I do think we could conditionalize the setup that uses the platforms somehow which i think would be all we'd have to do

test/rules/apple_verification_test.bzl
```diff
@@ -16,6 +16,8 @@
 
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 
+_supports_visionos = hasattr(apple_common.platform_type, "visionos")
```
brentleyjones, line 19: Nice!

brentleyjones (approved):

test/rules/apple_verification_test.bzl
keith: yea, easy one

PR #230, opened 2023-07-13
Remove cpp_linker_flags feature

keith:
We get `-lc++` from the `link_libc++` feature which is enabled for all
linking actions. We get `-target` from `default_link_flags` which is
enabled for all linking actions. This fixes ld-prime's new warning for
`cc_binary` targets and anything using `cc_common.link` (`swift_binary`).


keith:
related convo https://bazelbuild.slack.com/archives/C04DFUBQGSU/p1689257981712829

thii (approved):

PR #234, opened 2023-07-21
Remove -lc++ from non-C++ binary link actions

keith:
In the case you're linking an ObjC binary, we shouldn't add `-lc++` by
default, but users can still add this in their `sdk_dylibs` if they'd
like.


keith:
im not sure we should actually do this since i think we generally fixed the issue and this could break some folks who under-defined this dependency today

keith:
ha we even have that case in tests apparently

PR #238, opened 2023-07-31
Remove -lc++ from crosstool

keith:
As of Xcode 15 ld-prime warns about duplicate `-lfoo` flags. This
default flag results in a duplicate if anything in your dependency tree
sets `linkopts = ["-lc++"]` or `sdk_dylibs = ["c++"]`. Currently there
is no way for a crosstool to automatically add this flag and avoid this
duplication warning.

### DIFF
--- a/test/test_data/BUILD
+++ b/test/test_data/BUILD
@@ -17,12 +17,14 @@ TARGETS_UNDER_TEST_TAGS = [
 cc_binary(
     name = "cc_test_binary",
     srcs = ["main.cc"],
+    linkopts = ["-lc++"],
     tags = TARGETS_UNDER_TEST_TAGS,
 )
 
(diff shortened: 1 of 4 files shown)


keith:
all downsides right now, submitting this as a potential option. current workaround is to pass `--features=-link_libc++`

test/test_data/BUILD
```diff
@@ -17,12 +17,14 @@ TARGETS_UNDER_TEST_TAGS = [
 cc_binary(
     name = "cc_test_binary",
     srcs = ["main.cc"],
+    linkopts = ["-lc++"],
```
keith, line 20: this is super silly

keith:
we disabled this warning instead

PR #242, opened 2023-08-12
Remove apple_common transitions

keith:
This was removed upstream but we still need the smallest possible
transition for testing, this copies the one from rules_apple and removes
a bunch of the features. If this ends up being too much of a pain to
update we'll have to find a different solution.

### DIFF
--- a/test/starlark_apple_static_library.bzl
+++ b/test/starlark_apple_static_library.bzl
@@ -1,5 +1,7 @@
 """Test rule for static linking with bazel's builtin Apple logic"""
 
+load("//test:transitions.bzl", "apple_platform_split_transition")
+
 def _starlark_apple_static_library_impl(ctx):
     if not hasattr(apple_common.platform_type, ctx.attr.platform_type):
         fail('Unsupported platform type \"{}\"'.format(ctx.attr.platform_type))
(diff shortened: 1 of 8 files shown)


test/starlark_apple_static_library.bzl
```diff
@@ -71,16 +73,18 @@ starlark_apple_static_library = rule(
...
         "linkopts": attr.string_list(),
-        "platform_type": attr.string(),
-        "minimum_os_version": attr.string(),
+        "platform_type": attr.string(mandatory = True),
+        "minimum_os_version": attr.string(mandatory = True),
+        "_allowlist_function_transition": attr.label(
+            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+        ),
     },
-    cfg = apple_common.apple_crosstool_transition,
```
keith, line 84: not sure why but removing this seemed to be a no-op for our tests so i guess we don't have to worry about this one atm

test/transitions.bzl
```diff
@@ -0,0 +1,189 @@
...
+    "//command_line_option:ios_minimum_os",
+    "//command_line_option:macos_minimum_os",
+    "//command_line_option:platforms",
+    "//command_line_option:tvos_minimum_os",
+    "//command_line_option:watchos_minimum_os",
+]
+
+def _apple_platform_split_transition_impl(settings, attr):
+    output_dictionary = {}
+    if settings["//command_line_option:incompatible_enable_apple_toolchain_resolution"]:
```
keith, line 136: we don't need this branch for now but i figure we might later

PR #248, opened 2023-08-21
Move archive timestamp test to starlark

keith:
The only other test just tested that it built so we can drop that one


keith:
https://github.com/bazelbuild/apple_support/pull/250

PR #260, opened 2023-09-19
Return platforms from test transition

keith:


keith:
this is just to get the tests green, no impact on user behavior

brentleyjones (approved):

PR #262, opened 2023-09-19
Add platforms as input to universal transition

keith:
This way on Linux in the default case we forward the platforms


keith:
example failure https://buildkite.com/bazel/rules-swift-swift/builds/4555#018aaf34-f099-41ac-a65e-71bbafabd7b6

brentleyjones (approved):

PR #263, opened 2023-09-20
Add build flags to allow overriding platforms used in transitions

brentleyjones:
This allows people to use custom versions of the platforms, including child platforms of the default ones, which allows for custom additional constraints and platform properties.

### DIFF
--- a/configs/platforms.bzl
+++ b/configs/platforms.bzl
@@ -87,3 +87,23 @@ APPLE_PLATFORMS_CONSTRAINTS = {
         "@build_bazel_apple_support//constraints:simulator",
     ],
 }
+
+CPU_TO_DEFAULT_PLATFORM_NAME = {
+    "darwin_arm64": "macos_arm64",
+    "darwin_arm64e": "macos_arm64e",
+    "darwin_x86_64": "macos_x86_64",
+    "ios_arm64": "ios_arm64",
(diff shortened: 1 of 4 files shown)


thii (approved):

brentleyjones:
I now think they should have a suffix of `_default` instead of a prefix. Then it reads more like "this is cpu `ios_sim_arm`'s default platform".

And "default" is probably wrong, because the default value for them is the default, but you are mostly likely changing it away from the default if you use the flag.

So maybe it should actually be in a different package (root?) and named something like `@apple_support//:ios_sim_arm_platform` (which admittedly is a little confusing with the `platforms` package existing as well). This makes sense to me though because when set by a user they are expressing what is changing (`--@apple_support//:ios_sim_arm_platform=//custom_platforms:ios_sim` or their default value `--@apple_support//:ios_sim_arm_platform=@apple_support//platforms:ios_sim_arm`).

keith:
sounds fine, i also think it would be fine if they were still in the platforms package, but no strong pref

brentleyjones:
`--@apple_support//platforms:ios_sim_arm_platform=//custom_platforms:ios_sim`, yeah, that's not too bad, will go with that.

keith (approved):

PR #268, opened 2023-09-28
Migrate uses of ctx.fragments.apple.multi_arch_platform(...) to platform constraints.

thii:
PiperOrigin-RevId: 568558393
(cherry picked from commit 42d2c7cd00d57d9887cee68ecaa68d0029501fd6)

PR #269, opened 2023-09-28
Move methods to report platform constraints as strings to apple_support.

thii:
PiperOrigin-RevId: 568908396
(cherry picked from commit 4d4c409e744f1eb061bd0a93add789913430a9d2)

### DIFF
--- a/lib/apple_support.bzl
+++ b/lib/apple_support.bzl
@@ -460,6 +460,107 @@ def _run_shell(
         **kwargs
     ))
 
+def _target_arch_from_rule_ctx(ctx):
+    """Returns a `String` representing the target architecture based on constraints.
+
+    The returned `String` will represent a cpu architecture, such as `arm64` or `arm64e`.
+
+    In order to use `apple_support.target_arch_from_rule_ctx()`, you'll need to modify your rule
+    definition to add the following:
+
+      * Add the `apple_support.platform_constraint_attrs()` attributes to the `attrs` dictionary.
+        This can be done using the `dicts.add()` method from Skylib.
+
+    Args:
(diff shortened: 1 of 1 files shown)


lib/apple_support.bzl
```diff
@@ -460,6 +460,107 @@ def _run_shell(
...
+        A `String` representing the selected environment (e.g. `device`, `simulator`).
+    """
+    device_constraint = ctx.attr._apple_device_constraint[platform_common.ConstraintValueInfo]
+    simulator_constraint = ctx.attr._apple_simulator_constraint[platform_common.ConstraintValueInfo]
+
+    if ctx.target_platform_has_constraint(device_constraint):
+        return "device"
+    elif ctx.target_platform_has_constraint(simulator_constraint):
+        return "simulator"
+    fail("ERROR: A valid Apple environment (device, simulator) constraint could not be found from" +
```
brentleyjones, line 525: We relaxed this a bit in rules_apple.

thii: Added a similar change

brentleyjones: I do wonder now though, with what #268 is leading to, if we will need to tighten it back. Thoughts?

We only loosened it in rules_apple to allow people with out of date platforms to use rules_apple 3.0.0 more easily, but it seems like we will need these constraints properly setup for what rules_apple is going to want to do.

thii: Agree. If a platform is missing a `target_environment` constraint then the build would be broken half of the time already, since it would always assume it's targeting a "device" environment.

brentleyjones: Can you remove your extra commit then? I think it's good as original then.

PR #276, opened 2023-10-17
Upgrade stardoc to 0.6.2

brentleyjones:
Needed to have Bazel `last_green` tests pass.

PR #277, opened 2023-10-17
Disable `--features=treat_warnings_as_errors` on CI for Bazel 6

brentleyjones:
Bazel 6.x has dependencies that produce warnings now.


brentleyjones:
Requires #276, which requires this. Going to include this change in it.

PR #287, opened 2023-11-29
Add environ to repository rule

illicitonion:
Currently if you change this env var's value then re-run your build, the repository rule won't re-trigger so you'll get stale results.

Instead, make the repository rule re-trigger if the env var changes.


luispadron (approved):

keith:
thanks!

brentleyjones (approved):

PR #288, opened 2023-11-29
Only enable linkmap if the variable exists

keith:
In rules_apple we've started using cc_common.link in 1 place, which
doesn't set this variable in bazel since this is an Apple specific
feature in bazel. This means if you try to link a intents stub binary
and pass `--objc_generate_linkmap` the link will fail. We don't care
about generating linkmaps for that binary anyways since it's just an
intermediate artifact, so this fixes that link. This has the downside
that this could hide issues and silently not create linkmaps for links
that should support them in the future.


LoganShireSnapchat:
Thanks! I verified this fixes the issue locally.

keith:
thanks!

luispadron (approved):

PR #290, opened 2023-12-19
Update toolchain setup instructions

keith:
The big change here is that bazel 7.x shipped with toolchain resolution
flipped, so that no longer requires any configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,27 @@ building for Apple platforms besides macOS. Since Bazel 7 this toolchain
 is required when targeting those platforms but the toolchain also
 supports Bazel 6.
 
-To use the Apple CC toolchain, pull this repository into your build and
-add this to your `.bazelrc`:
+NOTE: This toolchain requires a full Xcode installation, not just the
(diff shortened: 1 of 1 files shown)


brentleyjones:
So we don't need this anymore for Bazel 7? What changed there?

```
build--apple_crosstool_top=@local_config_apple_cc//:toolchain
build--crosstool_top=@local_config_apple_cc//:toolchain
build--host_crosstool_top=@local_config_apple_cc//:toolchain
```

keith:
default flag flip for toolchain resolution

brentleyjones:
I thought that was unrelated to the `*crosstool_top` flags.

brentleyjones:
Should we mention that people need to set those still if they flip the flag off?

keith:
i left it out to try and reduce the instruction count making the assumption that those people are "on their own", especially since on rolling that flag is already a no-op, but I can add it if you think we should

brentleyjones (approved):

luispadron (approved):

PR #295, opened 2024-01-22
Enable visionOS tests always

keith:


keith:
not sure if CI supports it, we'll see, might need Xcode 15.2

luispadron (approved):

BalestraPatrick (approved):

Issue #298, opened 2024-01-28
Getting rid of apple_cc_toolchain


googlewalt:
We would like to get rid of `apple_cc_toolchain`.  There has been [recent efforts](https://github.com/bazelbuild/bazel/commit/ddcf089bd47c1888854081095cfa32fc1227fe3b) to refector it so that it is more like `cc_toolchain`, to the point where I believe the only remaining difference is the three crosstool variables used by `cc_toolchain_config` to set these environment variables: `XCODE_VERSION_OVERRIDE`, `APPLE_SDK_VERSION_OVERRIDE`, `APPLE_SDK_PLATFORM`.  This information is readily available in `xcode_config`, which is already a dependency of `cc_toolchain_config`, so I believe it should be straightforward to migrate the bazel apple cc toolchain to that.

Once that is done, we should be able to replace all instances of `apple_cc_toolchain` with `cc_toolchain`.

[This change](https://github.com/bazelbuild/bazel/commit/0af1cbae7c4e1b3a3fb9e141b08469e095901d2f) also removes the use on APPLE_SDK_VERSION_OVERRIDE, so that can eventually be dropped also.

keith:
first pass at this, there were a few more variables I had to reproduce, I assume once we merged this we could remove some of the logic from bazel's objc support https://github.com/bazelbuild/apple_support/pull/299

PR #299, opened 2024-01-29
Remove apple_cc_toolchain use

keith:
Fixes https://github.com/bazelbuild/apple_support/issues/298

Much of this logic was loosely copied from https://github.com/bazelbuild/bazel/blob/84dd8c129ba4b24a58f0159adc9c93286a02176f/src/main/starlark/builtins_bzl/common/objc/objc_common.bzl

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -45,12 +45,6 @@ _DYNAMIC_LINK_ACTIONS = [
     _OBJCPP_EXECUTABLE_ACTION_NAME,
 ]
 
-def _sdk_framework_dir(platform_type, sdk_version):
-    if platform_type == apple_common.platform_type.ios and sdk_version.compare_to(apple_common.dotted_version("9.0")) < 0:
-        return "__BAZEL_XCODE_SDKROOT__/Developer/Library/Frameworks"
-
-    return "__BAZEL_XCODE_SDKROOT__/System/Library/Frameworks"
-
 def _sdk_version_for_platform(xcode_config, platform_type):
     if platform_type == apple_common.platform_type.ios:
         return xcode_config.sdk_version_for_platform(apple_common.platform.ios_device)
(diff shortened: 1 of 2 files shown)


keith:
@googlewalt can you review?

crosstool/cc_toolchain_config.bzl
```diff
@@ -44,6 +45,48 @@ _DYNAMIC_LINK_ACTIONS = [
     _OBJCPP_EXECUTABLE_ACTION_NAME,
 ]
 
+def _sdk_framework_dir(platform_type, sdk_version):
+    if platform_type == apple_common.platform_type.ios and sdk_version.compare_to(apple_common.dotted_version("9.0")) < 0:
```
googlewalt, line 49: We shouldn't need this any more?

crosstool/cc_toolchain_config.bzl
```diff
@@ -1191,9 +1237,9 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
                     flag_group(
                         flags = [
                             "-isysroot",
-                            "%{sdk_dir}",
-                            "-F%{sdk_framework_dir}",
-                            "-F%{platform_developer_framework_dir}",
+                            "__BAZEL_XCODE_SDKROOT__",
+                            "-F{}".format(_sdk_framework_dir(platform_type, sdk_version)),
```
googlewalt, line 1241: FYI we use switched to using -iframework instead of -F internally, to silence some API availability warnings that were triggering in Xcode 15.0 XCTest headers.

googlewalt (changes requested):
Nice.  Thanks for the expediency!

crosstool/cc_toolchain_config.bzl
keith: ah yea, removed, note this is copied from bazel

crosstool/cc_toolchain_config.bzl
keith: interesting, we haven't gotten any reports about this yet, is that surprising?

googlewalt: I had to review the detail on the issue.  Apparently this doesn't happen on Xcode clang because it has some hack that suppresses diagnostics based on macro names (!).  So it shouldn't be an issue here.

googlewalt (approved):

PR #309, opened 2024-03-01
Allow disabling -dead_strip in opt builds

### LINKED ISSUE #308: Provide a mechanism to disable `-dead_strip` in builds
I've recently come across a failure to build gazelle that was ultimately traced down to commit a000b23a303edb7a614eba03cb4ea632225c8cde. I'm not a strong enough C++ developer to understand _why_ that's happening, but I noticed that the commit message for this stated that:

> This should be the right default for everyone, if not we can look at adding a way to disable it.

Having a flag to disable this feature would be most helpful.

keith:
Fixes https://github.com/bazelbuild/apple_support/issues/308

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -1059,11 +1059,6 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
                     ),
                 ],
             ),
-            flag_set(
-                actions = _DYNAMIC_LINK_ACTIONS,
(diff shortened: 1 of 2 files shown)


keith:
this is kinda weird but i think this is still the right default for most folks, and the risk of getting it wrong is too high wrt huge binary sizes

keith:
we should still do this

keith:
https://github.com/bazelbuild/apple_support/pull/626

PR #319, opened 2024-04-01
Exclude binary_macos_binary_with_spaces_test from BCR CI

luispadron:
Failing in: https://github.com/bazelbuild/bazel-central-registry/pull/1735

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -23,6 +23,7 @@ tasks:
     - '-@apple_support//test:linking_opt_link_test'
     - '-@apple_support//test:binary_watchos_device_arm64e_test'
     - '-@apple_support//test:binary_watchos_device_arm64_test'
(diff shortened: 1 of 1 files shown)


.bcr/presubmit.yml
```diff
@@ -23,6 +23,7 @@ tasks:
     - '-@apple_support//test:linking_opt_link_test'
     - '-@apple_support//test:binary_watchos_device_arm64e_test'
     - '-@apple_support//test:binary_watchos_device_arm64_test'
+    - '-@apple_support//test:binary_macos_binary_with_spaces_test'
```
brentleyjones, line 26: This wouldn't fix it, it's only failing in 7.x. I'm debugging.

luispadron: oh duh this list is for bazel 6, ill close this out

PR #321, opened 2024-04-01
Reapply "Add `gcc_quoting_for_param_files` feature (#314)"

brentleyjones:
This reverts commit cf271a330b08a3bbd8ad61241d03787683d5a1c5.

This adds a dependency on **bazel_features** in order to know when it's safe to enable `gcc_quoting_for_param_files` by default.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,6 +6,7 @@ module(
     repo_name = "build_bazel_apple_support",
 )
 
+bazel_dep(name = "bazel_features", version = "1.10.0")
 bazel_dep(name = "bazel_skylib", version = "1.3.0")
 bazel_dep(name = "platforms", version = "0.0.9")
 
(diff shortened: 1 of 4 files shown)


brentleyjones:
I'm not sure on the `apple_support_extra_dependencies` abstraction. Probably better to just have the `bazel_features_deps()` call listed in our snippet, similar to the **rules_xcodeproj** instructions. Also, it's technically a breaking change to add this for `WORKSPACE` users, but similar to how we handled it in **rules_swift**, we won't be upgrading the major version for this.

luispadron (approved):

mattrobmattrob (approved):

brentleyjones:
Now I just need to remember to add a note in the release notes.

keith:
I think you'll need to update this too: https://github.com/bazelbuild/apple_support/blob/master/.github/generate-notes.sh

brentleyjones:
https://github.com/bazelbuild/apple_support/pull/329

PR #325, opened 2024-04-23
Add support for layering_check

keith:
This feature currently applies to C/C++ compiles. It validates that for
every header you import you have a corresponding target in your deps
that surfaces that header. Bazel handles a lot of this logic internally
in the CC rules but it's up to us to provide a system-level modulemap
file that contains all system headers, so that non-system headers that
aren't in your deps are correctly flagged as a layering check violation.

To do this we use find to discover all headers in the current
DEVELOPER_DIR. There are a few potential caching issues surfaced with
this:

1. If the paths were absolute, the Xcode path would be an input. We
   avoid this by storing relative paths, and then using a vfsoverlay to
   make it "look" like the modulemap is inside of Xcode itself, which is
   the only way to make relative paths work in modulemaps.
2. If different Xcode versions have different headers, and you support
   multiple Xcode versions in your build, the modulemap file is an input
   and will cause caching issues. I think the future solution to this
   would be allow users to provide their own modulemap with a constant
   set of headers, since you likely won't care about the new or removed
   ones (and if you do you have to only support 1 of the Xcode verisons)

This is currently disabled by default and can be enabled with
`--repo_env=APPLE_SUPPORT_LAYERING_CHECK_BETA=1`. This is primarily
because of the issues above, and that generating the modulemap file
takes ~5 seconds.

### DIFF
--- a/crosstool/BUILD.tpl
+++ b/crosstool/BUILD.tpl
@@ -41,6 +41,13 @@ cc_toolchain_suite(
     toolchains = dict(CC_TOOLCHAINS),
 )
 
+filegroup(
+    name = "modulemap",
+    srcs = [
+%{real_modulemap}
+    ],
+)
+
 [
     filegroup(
         name = "osx_tools_" + arch,
@@ -49,6 +56,7 @@ cc_toolchain_suite(
             ":libtool",
             ":libtool_check_unique",
             ":make_hashed_objlist.py",
+            ":modulemap",
             ":wrapped_clang",
             ":wrapped_clang_pp",
             ":xcrunwrapper.sh",
@@ -71,6 +79,7 @@ cc_toolchain_suite(
         supports_param_files = 1,
         toolchain_config = arch,
         toolchain_identifier = arch,
+        module_map = %{placeholder_modulemap},
     )
     for arch in _APPLE_ARCHS
 ]
@@ -86,6 +95,7 @@ cc_toolchain_suite(
 %{cxx_builtin_include_directories}
         ],
         tool_paths_overrides = {%{tool_paths_overrides}},
+        module_map = ":modulemap",
     )
     for arch in _APPLE_ARCHS
 ]
--- a/crosstool/BUILD.tpl
+++ b/crosstool/BUILD.tpl
@@ -44,7 +44,7 @@ cc_toolchain_suite(
 filegroup(
     name = "modulemap",
     srcs = [
-%{real_modulemap}
+%{layering_check_modulemap}
     ],
 )
 
--- a/crosstool/generate-modulemap.sh
+++ b/crosstool/generate-modulemap.sh
@@ -6,7 +6,7 @@ cd "$(xcode-select -p)"
 
 echo 'module "crosstool" [system] {'
 
-find .. -type f \( -name "*.h" -o -name "*.def" -o -path "*/c++/*" \) \
+find . -type f \( -name "*.h" -o -name "*.def" -o -path "*/c++/*" \) \
   | LANG=C sort -u | while read -r header; do
     echo "  textual header \"${header}\""
   done
--- a/crosstool/generate-modulemap.sh
+++ b/crosstool/generate-modulemap.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cd "$(xcode-select -p)"/Platforms
+
+echo 'module "crosstool" [system] {'
+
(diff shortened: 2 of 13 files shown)


brentleyjones:
> If different Xcode versions have different headers, and you support
multiple Xcode versions in your build, the modulemap file is an input
and will cause caching issues. I think the future solution to this
would be allow users to provide their own modulemap with a constant
set of headers, since you likely won't care about the new or removed
ones (and if you do you have to only support 1 of the Xcode verisons)

Since you can transition on `--xcode_version`, this is very easy to do (as in, very easy to get into the state that you are building with multiple Xcode versions).

keith:
Yea I think it would make sense to allow users to provide their config to support that, I just didn't want to do it in the initial testing. I don't see a great way for us to automatically solve that, since if we generated this file for all the xcode versions users had installed, the combination of installed Xcode versions would become an input

crosstool/BUILD.tpl
```diff
@@ -41,6 +41,13 @@ cc_toolchain_suite(
     toolchains = dict(CC_TOOLCHAINS),
 )
 
+filegroup(
+    name = "modulemap",
+    srcs = [
+%{real_modulemap}
+    ],
+)
```
brentleyjones, line 49: Why do you go through a `filegroup` instead of referencing it directly?

crosstool/osx_cc_configure.bzl
```diff
@@ -207,6 +242,8 @@ def configure_osx_toolchain(repository_ctx):
             "%{tool_paths_overrides}": ",\n            ".join(
                 ['"%s": "%s"' % (k, v) for k, v in tool_paths.items()],
             ),
+            "%{real_modulemap}": "\":{}\",".format(real_modulemap) if real_modulemap else "",
+            "%{placeholder_modulemap}": "\":module.modulemap\"" if real_modulemap else "None",
```
brentleyjones, line 246: nit: can you make these alphabetical?

crosstool/setup.bzl
```diff
@@ -57,6 +57,7 @@ _apple_cc_autoconf = repository_rule(
         _DISABLE_ENV_VAR,
         _OLD_DISABLE_ENV_VAR,
         "BAZEL_ALLOW_NON_APPLICATIONS_XCODE",  # Signals to configure_osx_toolchain that some Xcodes may live outside of /Applications and we need to probe further when detecting/configuring them.
+        "APPLE_SUPPORT_LAYERING_CHECK_BETA",
```
brentleyjones, line 60: nit: Can you move this up to be alphabetical?

crosstool/BUILD.tpl
keith: otherwise we'd need 2 different substitutions because for the osx_tools target we would have to substitute to nothing, and for the cc_toolchain_config target we'd have to either substitute to `None` or `module_map = thing`

crosstool/generate-modulemap.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cd "$(xcode-select -p)"
+
+echo 'module "crosstool" [system] {'
```
luispadron, line 7: Do you know if quotes are needed here? The examples over at: https://clang.llvm.org/docs/Modules.html#module-maps don't use quotes and there was some discussion in the slack recently about this causing issues.

crosstool/osx_cc_configure.bzl
```diff
@@ -181,6 +207,15 @@ def configure_osx_toolchain(repository_ctx):
     _compile_cc_file(repository_ctx, wrapped_clang_src_path, "wrapped_clang")
     repository_ctx.symlink("wrapped_clang", "wrapped_clang_pp")
 
+    real_modulemap = None
+    if repository_ctx.os.environ.get("APPLE_SUPPORT_LAYERING_CHECK_BETA") == "1":
+        real_modulemap = "real.modulemap"
```
luispadron, line 212: Would having layering_check in the name be more obvious here if this is only used when that's enabled?

luispadron (approved):

crosstool/generate-modulemap.sh
keith: doesn't seem to make a difference, I was just doing what bazel is doing for linux:

https://github.com/bazelbuild/bazel/blob/5ebdbfce9134f67bae664b900b5d53b215a4fcce/tools/cpp/generate_system_module_map.sh#L18

we can revisit if it breaks someone, I assume that thread is different somehow

crosstool/osx_cc_configure.bzl
keith: renamed

PR #326, opened 2024-04-26
Enable supports_dynamic_linker feature on darwin

fffonion:
The `supports_dynamic_linker` is not turned on, resulting shared library is not producde by cc_library or cc_test

Ref: https://github.com/bazelbuild/bazel/issues/16479 https://github.com/bazelbuild/bazel/blob/c5cb07ac4603cef03d7037d603971e1f4e9184c9/src/main/starlark/builtins_bzl/common/cc/cc_library.bzl#L94

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -2604,6 +2604,7 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
         ctx.attr.cpu == "darwin_arm64" or
         ctx.attr.cpu == "darwin_arm64e"):
         features.append(feature(name = "dynamic_linking_mode"))
+        features.append(feature(name = "supports_dynamic_linker"))
 
     # macOS artifact name patterns differ from the defaults only for dynamic
     # libraries.


fffonion:
The linked ref also mentioned it was removed due to https://github.com/bazelbuild/bazel/commit/ec5553352f2f661d39ac4cf665dd9b3c779e614c. But I'm not sure if it's safe to enable just on darwin.

keith:
It's not safe to. For the same reason from that commit message. There are some other threads about this around too I can probably dig up later

fffonion:
Thanks for the reply @keith . In that case is `cc_shared_library` the preferred way to get a `.dylib` on macOS? I'm bit hesitated
to use it as it's marked as experimental.

fffonion:
Thank you!

fffonion:
I see, thanks for you detailed reply. I now understands the limitation on macOS. I do have a generic question regarding those cc_* rules after reading your reply though. If I want to produce a shared library for binary redistribution, for example I want to submit libmyproject to homebrew. Is it true that I would better use cc_binary or cc_shared_library instead of cc_library, as the later is intended for intermediate usage? The reason I'm asking this is because it seems `cc_binary` doesn't export any functions in the produced shared library by default, but cc_shared_library does. So I was wondering if I was going the wrong path, as I see the `.so` file produced by cc_library on linux does export functions so it can used for both linking and distribution.

keith:
i think you'd be better off with `cc_shared_library`, i wouldn't be worried about the experimental-ness anymore (with bazel 7 it's no longer named experimental). I think you could make it work with cc_binary though with -exported_symbols_list on macOS

fffonion:
Thank you again! I will have a try of that flag!

PR #328, opened 2024-05-16
Mark module extension as reproducible

fmeum:
This removes an unnecessary lockfile entry.

The module extension is kept in place so that `use_extension` calls don't need to be updated. The repo rules are moved into a separate file so that WORKSPACE setups don't fail on a cyclic load of `bazel_features`. The newly exposed repo rules are protected by load visibility.

### DIFF
--- a/crosstool/setup.bzl
+++ b/crosstool/setup.bzl
@@ -1,82 +1,14 @@
 """Configure the Apple CC toolchain"""
 
-load("//crosstool:osx_cc_configure.bzl", "configure_osx_toolchain")
+load("@bazel_features//:features.bzl", "bazel_features")
+load(":setup_internal.bzl", "apple_cc_autoconf", "apple_cc_autoconf_toolchains")
 
-_DISABLE_ENV_VAR = "BAZEL_NO_APPLE_CPP_TOOLCHAIN"
-_OLD_DISABLE_ENV_VAR = "BAZEL_USE_CPP_ONLY_TOOLCHAIN"
-
-def _apple_cc_autoconf_toolchains_impl(repository_ctx):
-    """Generate BUILD file with 'toolchain' targets for the local host C++ toolchain.
-
-    Args:
-      repository_ctx: repository context
-    """
-    env = repository_ctx.os.environ
-    should_disable = _DISABLE_ENV_VAR in env and env[_DISABLE_ENV_VAR] == "1"
-    old_should_disable = _OLD_DISABLE_ENV_VAR in env and env[_OLD_DISABLE_ENV_VAR] == "1"
-
-    if should_disable or old_should_disable:
-        repository_ctx.file("BUILD", "# Apple CC toolchain autoconfiguration was disabled by {} env variable.".format(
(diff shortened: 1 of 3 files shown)


crosstool/setup.bzl
```diff
@@ -27,7 +27,7 @@ def _apple_cc_autoconf_toolchains_impl(repository_ctx):
     else:
         repository_ctx.file("BUILD", "# Apple CC toolchain autoconfiguration was disabled because you're not running on macOS")
 
-_apple_cc_autoconf_toolchains = repository_rule(
+apple_cc_autoconf_toolchains = repository_rule(
```
fmeum, line 30: If it's a concern that these symbols are now visible, I could restructure the file further and adopt `visibility()` to prevent that.

brentleyjones: I would prefer not to increase visibility on any of our symbols.

fmeum: I rearranged the contents so that I can keep the symbols private. This is assuming that your minimum supported version of Bazel is 6 (for load visibility).

brentleyjones: Currently our minimum is Bazel 5.

fmeum: I see, then I don't see a way to enforce visibility. Is the name `setup_internal.bzl` good enough or should I move the file to some `internal` or `private` subdirectory?

brentleyjones: Under an `internal/` directory will add some buildifier enforcement as well, so that should be good enough.

fmeum: I moved it.

MODULE.bazel
```diff
@@ -17,7 +17,7 @@ bazel_dep(
     repo_name = "io_bazel_stardoc",
 )
 
-apple_cc_configure = use_extension("//crosstool:setup.bzl", "apple_cc_configure_extension")
+apple_cc_configure = use_extension("//crosstool:apple_cc_configure_extension.bzl", "apple_cc_configure_extension")
```
brentleyjones, line 20: Is this file rename a breaking change for users? We have a blurb in the README that references the old name.

crosstool/internal/BUILD.bazel
brentleyjones, line 1: In this repo we use `BUILD` instead of `BUILD.bazel` if possible.

MODULE.bazel
fmeum: It would be, yes. I think it's better to wait for apple_support to drop Bazel 5 support and then land this commit without the need for this breaking change.

luispadron (approved):

crosstool/setup.bzl
```diff
@@ -1,5 +1,6 @@
 """Configure the Apple CC toolchain"""
 
+load("@bazel_features//:features.bzl", "bazel_features")
```
luispadron, line 3: We'll need to add this to `MODULE.bazel`

luispadron: Or i guess its complaining in `WORKSPACE.bazel` when not using bzlmod, since its already in the bzlmod file.

PR #335, opened 2024-06-07
Remove `apply_implied_frameworks` feature from `implies` action configs

### LINKED ISSUE #334: Provide a way to disable implicit framework linking
Reporting here from a conversation in Slack.

The rules_apple [docs for sdk_frameworks](https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-apple.md) say that UIKit will always be linked if targeting iOS etc. I believe I've traced this down to the `apply_implicit_frameworks` feature defined here in apple_support.

For File Provider extensions, the available memory limit is very low (roughly 15MB on iOS 17), so we have some checks in place in our Buck build to prevent UIKit from getting linked into that extension. We already have implicit framework linking behavior disabled in Buck. We're moving our build from Buck to Bazel and would like to retain this behavior.

It was already explained to me that disabling implied features is not a straightforward process without customization of the ruleset. Would it be possible or reasonable to create an escape hatch for this feature?

aaronsky:
Fixes #334

### DIFF
--- a/test/linking_tests.bzl
+++ b/test/linking_tests.bzl
@@ -42,76 +42,92 @@ def linking_test_suite(name):
     Args:
         name: The name to be included in test names and tags.
     """
-    default_test(
-        name = "{}_default_apple_link_test".format(name),
-        tags = [name],
-        expected_argv = [
-            "-Xlinker",
-            "-objc_abi_version",
-            "-Xlinker",
-            "2",
-            "-ObjC",
-        ],
-        not_expected_argv = [
-            "-g",
-            "DSYM_HINT_LINKED_BINARY",
-            "-dead_strip",
-        ],
-        mnemonic = "ObjcLink",
-        target_under_test = "//test/test_data:macos_binary",
-    )
+    framework_flags = {
+        "macos": ["-framework", "Foundation"],
+        "ios": ["-framework", "Foundation", "-framework", "UIKit"],
+    }
 
-    opt_test(
-        name = "{}_opt_link_test".format(name),
-        tags = [name],
-        expected_argv = [
-            "-Xlinker",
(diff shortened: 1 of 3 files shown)


keith:
nice, can you add a test that verifies they are still passed by default and are not passed when disabled? https://github.com/bazelbuild/apple_support/blob/288cfc0646781abab6ce4372e34d335ad9245768/test/linking_tests.bzl

test/linking_tests.bzl
```diff
@@ -42,76 +42,92 @@ def linking_test_suite(name):
...
-            "-objc_abi_version",
-            "-Xlinker",
-            "2",
-            "-ObjC",
-            "-dead_strip",
-        ],
-        mnemonic = "ObjcLink",
-        target_under_test = "//test/test_data:macos_binary",
-    )
+    for platform in ["macos", "ios"]:
```
keith, line 50: i think this is probably overkill, you can probably just add the framework to the default case, and then create a new rule at the top of the file to transition the features off, and assert that without any new targets and without adding it to each test

aaronsky: I'd be happy to do that, do you know of any good examples of how to transition on a feature? This is an area of Bazel I have not yet developed muscles for yet.

keith: https://github.com/bazelbuild/apple_support/blob/288cfc0646781abab6ce4372e34d335ad9245768/test/linking_tests.bzl#L19-L21

i just mean with this test infra

aaronsky: that makes sense, thank you!

luispadron (approved):

brentleyjones (approved):

keith:
nice!

PR #336, opened 2024-06-19
Allows to provide Lipo tool location via custom toolchain.

DmitriiTcimbal:


keith:
I think im wrong about that and it must be a no-op for linux at the very least. at first i wasn't thinking we should merge this but i've since seen that grpc started using this rule, and if you want to avoid major xcode version dependence you need this, so i think we should probably move forward with it. can you rebase and sign the CLA?

PR #341, opened 2024-08-12
support selectively disabling NS_BLOCK_ASSERTIONS for opt builds

aaronsky:
maintains the default behavior of passing `-DNS_BLOCK_ASSERTIONS=1` for opt builds. for builds where this is undesirable, set `--features=-ns_block_assertions` to unconditionally not-set this flag, without having to unset the entire `default_compile_flags` feature.

does not add tests, but i can add them if requested.

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -2077,7 +2077,6 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
                             "-g0",
                             "-O2",
                             "-DNDEBUG",
-                            "-DNS_BLOCK_ASSERTIONS=1",
                         ],
                     ),
                 ],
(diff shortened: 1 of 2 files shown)


brentleyjones (approved):

luispadron (approved):

keith:
previously this was always enabled, for dbg/fastbuild/opt, now it looks like it's only enabled for opt? i think tests would be nice to cover the behavior here

aaronsky:
@keith From what I can tell, that isn't the case. My understanding is that `-DNS_BLOCK_ASSERTIONS=1` is only being set in opt builds. I could easily be misinterpreting the existing code, but I was able to confirm that understanding in my own builds.

I am willing to write tests, but was having trouble finding any existing clang flag tests. Do you know where I could find them?

keith:
oh you're right, the constraint was folded below more lines than i expected. here's a file you can add tests in https://github.com/bazelbuild/apple_support/pull/342

brentleyjones (approved):

keith:
thanks!

PR #343, opened 2024-09-04
Use `-Xlinker` for `-install_name`

fmeum:
This supports `,` in `runtime_solib_name`.

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -2439,7 +2439,10 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
                 flag_groups = [
                     flag_group(
                         flags = [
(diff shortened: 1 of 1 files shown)


fmeum:
Would it be possible to get a release after this has been merged? This would unblock enabling C++ remote execution tests for macOS in Bazel (https://github.com/bazelbuild/bazel/pull/23085).

fmeum:
Target names with commas can trigger this (there's a Bazel test for this case). The feature above this one already uses `-Xlinker`. If deduplication doesn't work, that sounds like a bug we should solve regardless.

keith (approved):

keith:
fwiw we tried to fix that but weren't able to land a pr upstream

fmeum:
Can you point me to the PR?

keith:
see the deduplication stuff in this issue https://github.com/bazelbuild/bazel/issues/16939#issue-1480421627

PR #348, opened 2024-11-07
Enable link time -Os in opt builds

cerisier:
`ld` starting Xcode 15 (`ld_prime`) now supports this option to perform more general code de-deduplication passes which it was not doing prior.

It is not documented anywhere but Xcode 15+ has this enabled by default in the default Release scheme (which has`Optimization Level = Fastest/Smallest -Os` by default).

I'm adding this as a separate feature so that it can be disabled but there could be an argument whether to enable this by default !

Fun note: passing any character after `-O` performs the same as `-Os`...

### DIFF
--- a/test/linking_tests.bzl
+++ b/test/linking_tests.bzl
@@ -44,6 +44,15 @@ dsym_test = make_action_command_line_test_rule(
     },
 )
 
+opt_osize_enabled_test = make_action_command_line_test_rule(
+    config_settings = {
+        "//command_line_option:compilation_mode": "opt",
+        "//command_line_option:features": [
+            "link_osize_opt",
+        ],
+    },
+)
+
 def linking_test_suite(name):
     """Tests for linking behavior.
 
@@ -118,6 +127,22 @@ def linking_test_suite(name):
             "2",
             "-ObjC",
             "-dead_strip",
+        ],
+        mnemonic = "ObjcLink",
+        target_under_test = "//test/test_data:macos_binary",
+    )
+
+    opt_osize_enabled_test(
+        name = "{}_opt_osize_enabled_test".format(name),
+        tags = [name],
+        expected_argv = [
+            "-Xlinker",
+            "-objc_abi_version",
+            "-Xlinker",
+            "2",
+            "-ObjC",
+            "-Xlinker",
+            "-Os",
         ],
         mnemonic = "ObjcLink",
         target_under_test = "//test/test_data:macos_binary",
--- a/test/linking_tests.bzl
+++ b/test/linking_tests.bzl
@@ -53,6 +53,15 @@ opt_osize_enabled_test = make_action_command_line_test_rule(
     },
 )
 
+disable_opt_osize_test = make_action_command_line_test_rule(
+    config_settings = {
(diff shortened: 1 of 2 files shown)


test/linking_tests.bzl
```diff
@@ -123,6 +132,22 @@ def linking_test_suite(name):
         target_under_test = "//test/test_data:macos_binary",
     )
 
+    opt_osize_enabled_test(
```
brentleyjones, line 135: Can you also add a test that disables the feature?

cerisier: Done

test/linking_tests.bzl
```diff
@@ -123,6 +141,36 @@ def linking_test_suite(name):
...
+        name = "{}_opt_osize_enabled_test".format(name),
+        tags = [name],
+        expected_argv = [
+            "-Xlinker",
+            "-objc_abi_version",
+            "-Xlinker",
+            "2",
+            "-ObjC",
```
brentleyjones, line 152: This part isn't needed.

test/linking_tests.bzl
```diff
@@ -123,6 +141,36 @@ def linking_test_suite(name):
...
+
+    disable_opt_osize_test(
+        name = "{}_disable_opt_osize_test".format(name),
+        tags = [name],
+        expected_argv = [
+            "-Xlinker",
+            "-objc_abi_version",
+            "-Xlinker",
+            "2",
+        ],
```
brentleyjones, line 168: This part isn't needed.

brentleyjones (approved):

fhanau:
@cerisier Can you provide an example where you see an effect on binary size from using the flag? I'm using Xcode 16 and it looks to me like it is on-by-default – when I add just `-Wl,-Os` there is no effect on binary size. If I pass `-Wl,-O0` followed by `-Wl,-Os` the output has the same size as when no flags are passed, suggesting that -Os reverts the increase in binary size that would otherwise be caused by -O0 but doesn't have an effect just by itself.

Additionally this breaks the build when using lld as the linker (for example when having Homebrew lld installed and using `--linkopt="-fuse-ld=lld"`) – I know that apple_support does not explicitly support lld but it would be nice to not have a regression for it.

brentleyjones:
When using lld you can disable the feature. Per the PR description, passing any `-O` to the linker applies this, so you can't undo it in the Xcode build.

fhanau:
> Per the PR description, passing any `-O` to the linker applies this, so you can't undo it in the Xcode build.

The PR description is incorrect then – passing `-O0` reverts the effect of using -Os or a different -O option for me locally.
Furthermore, the behavior of -Os or similar appears to be the default behavior – to be specific I'm seeing this when invoking the linker via the clang++ driver as opposed to via Xcode directly. If `-Wl,-Os` only specifies the default behavior, then there is no need to pass this flag in opt mode. Perhaps builds through Xcode default to the -O0 behavior in some configurations?

jpsim:
Heads up that I'm seeing some non-hermeticity with this change: https://github.com/bazelbuild/apple_support/issues/371

PR #352, opened 2024-12-06
Drop support for bazel 6

aaronsky:
Matching what was done in bazelbuild/rules_apple#2610


keith:
change looks good, but def an interesting policy question since this repo is potentially used for non rules_apple projects

brentleyjones:
In **rules_apple** we needed to do this to support the new linking stuff. I don't think we need to do this yet. We should feel fine with dropping LTS-2 if we need to, but I don't think we gain much by doing it in **apple_support** right now?

keith:
@aaronsky 6.x has gotten more annoying to support, want to reopen this so we can move forward with it?

keith (approved):

aaronsky:
@keith with this test failure, i know your libtool change slightly altered the hashing representation of objects – could that be related?

keith:
i don't see what in this PR would change that yet, but it's fine to update the tests w/ the new values

luispadron:
Nice this makes the other cherry picks easier to land, I'll get that Xcode rules fork PR updated in a bit

PR #356, opened 2024-12-20
Add no_dotd_file well known feature

keith:
This feature is used in bazel to disable anything related to .d files.
This is off by default but must be here to be enabled.


luispadron (approved):
Could this be related to https://github.com/bazelbuild/bazel/issues/22387 at all?

keith:
only in that it would likely sidestep that all together, because the related actions shouldn't ever be registered

aaronsky (approved):

brentleyjones (approved):

keith:
waiting on https://github.com/bazelbuild/rules_cc/pull/296

PR #357, opened 2025-01-14
Copy osx_cc_wrapper.sh.tpl from bazel

keith:
This is no longer used in bazel so we should just own it. This copies
over the relevant shell test too.


luispadron (approved):

aaronsky (approved):

keith:
macOS CI very borked

PR #361, opened 2025-02-04
Use build rules for some toolchain tools

brentleyjones:
This causes the tools to be built on the execution platform, which might not be the same as the host platform. This is needed to support cross-platform RBE builds.

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -249,7 +249,7 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
         ],
         tools = [
             tool(
-                path = "wrapped_clang",
+                tool = ctx.file.wrapped_clang,
                 execution_requirements = xcode_execution_requirements,
             ),
         ],
@@ -282,7 +282,7 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
         ],
         tools = [
             tool(
-                path = "wrapped_clang",
+                tool = ctx.file.wrapped_clang,
                 execution_requirements = xcode_execution_requirements,
             ),
         ],
@@ -348,7 +348,7 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
         ],
         tools = [
             tool(
-                path = "wrapped_clang",
+                tool = ctx.file.wrapped_clang,
                 execution_requirements = xcode_execution_requirements,
             ),
         ],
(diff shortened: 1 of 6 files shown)


crosstool/cc_toolchain_config.bzl
```diff
@@ -2774,10 +2774,10 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
     ]
 
     tool_paths = {
-        "ar": "libtool",
+        "ar": ctx.file.libtool.path,
         "cpp": "/usr/bin/cpp",
         "dwp": "/usr/bin/dwp",
-        "gcc": "cc_wrapper.sh",
+        "gcc": ctx.file.cc_wrapper.path,
```
brentleyjones, line 2780: I think there is a different way I should be doing this.

Related, I think: https://github.com/bazelbuild/bazel/issues/8438#issuecomment-1523519395

brentleyjones: Yeah we cover those with our current `action_config`, so I can remove these two.

brentleyjones: Hmm, dang:
```
(22:23:53) ERROR: /private/var/tmp/_bazel_buildkite/eb7275b476ff60c9bee15cac7036007d/external/local_config_apple_cc/BUILD:89:19: in cc_toolchain_suite rule @local_config_apple_cc//:toolchain: Tool path for 'ar' is missing
```

There is probably something else I can do though.

crosstool/universal_exec_tool.bzl
```diff
@@ -0,0 +1,30 @@
+def universal_exec_tool(*, name, out, srcs):
+    native.genrule(
+        name = name,
```
keith, line 3: I think you actually want apple_genrule so we get the Xcode variables?

brentleyjones: Pretty sure that makes a circular reference. I'll try it though.

brentleyjones: Hmm, it worked, sweet!

crosstool/cc_toolchain_config.bzl
brentleyjones: Looks like some rules directly ask for `"ar"`, so we need this.

keith (approved):

brentleyjones:
I'll need to make Buildifier green in another PR.

PR #364, opened 2025-02-05
Enable the `archive_param_file` feature by default

brentleyjones:
Matches rules_cc: https://github.com/bazelbuild/rules_cc/blob/6a2520bed08bbe47a8afcf4c43d51c2533d79ed8/cc/private/toolchain/unix_cc_toolchain_config.bzl#L1648.

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -2664,7 +2664,7 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
 
     features = [
         # Marker features
-        feature(name = "archive_param_file", enabled = True),
+        feature(name = "archive_param_file"),
         feature(name = "compile_all_modules"),
         feature(name = "coverage"),
         feature(name = "dbg"),


crosstool/cc_toolchain_config.bzl
```diff
@@ -2664,7 +2664,7 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
 
     features = [
         # Marker features
-        feature(name = "archive_param_file", enabled = True),
+        feature(name = "archive_param_file"),
```
luispadron, line 2667: The rules_cc one is using `enabled = True` but this removes it, just checking that was expected

keith: Yea looks like this did the opposite of the description?

brentleyjones: Ohh, dang. Let me revert this. Not sure where the mixup came.

adincebic: Was a little confused because of https://github.com/bazelbuild/apple_support/pull/338

PR #366, opened 2025-02-07
Use a BCR test module for tests

brentleyjones:
Needed to use dev dependencies in BCR presubmit.

https://buildkite.com/bazel/bcr-presubmit/builds/10562

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -8,15 +8,21 @@ tasks:
     bazel: ${{ bazel }}
     build_targets:
     - '@apple_support//lib/...'
-  run_tests:
-    name: "Run test targets"
-    platform: "macos_arm64"
-    bazel: ${{ bazel }}
-    build_targets:
-    - '@apple_support//lib/...'
-    test_targets:
-    - '@apple_support//test/...'
-    - '--'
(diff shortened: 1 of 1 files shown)


.bcr/presubmit.yml
```diff
@@ -8,15 +8,21 @@ tasks:
...
-    - '--'
-    # Needs visionOS SDK
-    - '-@apple_support//test:binary_visionos_arm64_simulator_test'
-    - '-@apple_support//test:binary_visionos_device_test'
+
+bcr_test_module:
+  module_path: "."
+
+  matrix:
+    bazel: ["7.x", "rolling"]
```
mattrobmattrob, line 16: This isn't already inherited from the top-level declaration, right?

mattrobmattrob (approved):

.bcr/presubmit.yml
brentleyjones: I don't believe so based on the docs.

aaronsky (approved):

.bcr/presubmit.yml
brentleyjones: Confirmed it's needed.

PR #367, opened 2025-02-09
add `suppress_warnings` feature

aaronsky:
This is something I could use in my codebase. I would like have a way to propagate the `-w` flag to all clang-using targets, and specifically, selectively disable it in my REPO.bazel and individual targets. This would allow me to globally suppress warnings for external repositories, without enabling it in my own code, and without needing to juggle the state of the flag setting (`-w` and other diagnostic flags are mutually exclusive afaik). This is the same as how we organize the `treat_warnings_as_errors` feature today. 

Companion change in rules_swift is bazelbuild/rules_swift#1489

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -2564,6 +2564,21 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
         ],
     )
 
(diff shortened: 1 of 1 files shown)


luispadron (approved):

aaronsky:
yup, agreed!

PR #368, opened 2025-02-15
Enable tests that require bazel 8.x

keith:


keith:
ah ok so it means once we drop others

PR #373, opened 2025-02-19
Remove no_uuid and codesigning workarounds when building wrapped_clang.

mkosiba:
The core problem was fixed in https://reviews.llvm.org/D111269 and these
workarounds are no longer necessary.

### DIFF
--- a/crosstool/universal_exec_tool.bzl
+++ b/crosstool/universal_exec_tool.bzl
@@ -38,18 +38,9 @@ env -i \
     -lc++ \
     -arch arm64 \
     -arch x86_64 \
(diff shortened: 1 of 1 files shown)


keith:
If we remove this I think we should remove the codesigning workaround too

keith (approved):

mkosiba:
thanks @keith! I think you (or another contributor) will have to merge the PR on my behalf as well as I'm not authorized.

keith:
yea np I was just giving it some time in case anyone else wanted to review

keith:
thanks!

PR #376, opened 2025-03-14
Update C++ standard default to 17

aaronsky:
In accordance with Google's [Foundational C++ Support matrix](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md), and given other Google projects have moved to adopt C++17 as a minimum or default ([protobuf](https://github.com/protocolbuffers/protobuf/commit/fe535930d3183b46d7e088683cbe8c49285715f8), [rules_cc](https://github.com/search?q=repo%3Abazelbuild%2Frules_cc+%22-std%22&type=code)), I'd like us to consider bumping this version in apple_support as well.

This constitutes a breaking change. It uses the change made in #219 as a reference.

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -447,7 +447,7 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
                             "-target",
                             target_system_name,
                             "-stdlib=libc++",
-                            "-std=gnu++14",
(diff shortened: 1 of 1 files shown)


mattrobmattrob (approved):
Seems reasonable to me. Just want to organize any other major version bump changes if/when we merge this.

keith:
best reason to do this is that the default C++ toolchain did as well :+1:

PR #377, opened 2025-03-20
Make it possible to use llvm-cov with `bazel coverage`

mkosiba:
The fact that `llvm-cov` and `llvm-profdata` are not set seems to trip up the `external/bazel_tools/tools/test/collect_cc_coverage.sh` script resulting in empty .dat files.

To solve the issue we implement something similar to what's done in [rules_cc](https://github.com/bazelbuild/rules_cc/blob/d74915024017250e46d95e91a3defc34174effe0/cc/private/toolchain/unix_cc_configure.bzl#L369). 

After these changes `bazel coverage <test target>` no longer produces empty .dat files.


crosstool/osx_cc_configure.bzl
```diff
@@ -88,6 +88,26 @@ def _generate_system_modulemap(repository_ctx, script, output):
...
+            "env",
+            "-i",
+            "DEVELOPER_DIR={}".format(env.get("DEVELOPER_DIR", default = "")),
+            "xcrun",
+            "--find",
+            tool_name,
+        ])
+        if result.return_code != 0:
+            return None
+        return result.stdout.strip()
```
keith, line 109: does these being host specific absolute paths affect caching? since this will have the absolute path to Xcode.app in it

keith: Yea i think that would work.

I think the reason the gcov one has flown under the radar is because in apple specific (aka non-c++) tests the coverage logic is handled in rules_apple and sidesteps all of this

mkosiba: no worries. I think the updated version addresses this issue. If no overrides are present then the cache key contains a script which just says `xcrun --run gcov/llvm-profdata/llvm-cov`. Since the script is the same across machines caching should work fine.
If there are override env vars present, then as long as users set the override to the same value, they should also get cache hits.

PR #379, opened 2025-03-26
Remove _xcrunwrapper

keith:
This used to be used internally by bazel


keith:
I think bazel 6.x still needs this, want to see if CI catches it

keith:
https://github.com/bazelbuild/apple_support/pull/380

PR #384, opened 2025-03-27
Add test for archive file hashes

keith:
This makes sure we don't break our libtool behavior to hash files if
there are duplicate basenames


brentleyjones (approved):

keith:
this is a bit fragile since this hash contains the transition hash, we'll see how annoying that is

PR #388, opened 2025-03-27
Replace libtool shell template with C++

keith:
This combines libtool.sh.tpl, libtool_check_unique.cc, xcrunwrapper.sh,
and make_hashed_objlist.py into a single C++ driver.

This sidesteps issues with our toolchain tools having to discover
each other, and reduces some duplication in this setup.


adincebic:
Is there a reason why not use Swift and SwiftArgumentParser instead of cpp?

keith:
There's a bit of a circular dependency problem as this file is required to produce any swift_library. In general the toolchain files are much easier to manage if they're single files. For just swift with no deps vs C++ the downside of using swift is that we're likely more susceptible to having to fight user's Xcode versions.

aaronsky (approved):

PR #390, opened 2025-03-27
Delete osx_cc_wrapper.sh

keith:
This was only used to fix the install_name of shared libraries, which is
now obsolete since they are set correctly by the crosstool with
`-install_name`. This was fixed in 7.4.0+ https://github.com/bazelbuild/bazel/commit/4956dff50d21a3346c1a63985910df10fccfd90b


keith:
should see on CI that it's still required for bazel before 8.x

aaronsky (approved):

brentleyjones:
Looks like https://github.com/bazelbuild/bazel/commit/4956dff50d21a3346c1a63985910df10fccfd90b is what fixed it. Not sure we should move our minimum up to 7.4.0 just yet?

keith:
probably fine to make 7.4.0 our min now?

adincebic (approved):

adincebic:
> probably fine to make 7.4.0 our min now?

IMO yes.

PR #394, opened 2025-03-28
Replace cpu config_setting with constraint_values

keith:
Using a config_setting based on cpu is deprecated

### DIFF
--- a/configs/BUILD
+++ b/configs/BUILD
@@ -15,9 +15,9 @@ selects.config_setting_group(
 [
     config_setting(
         name = cpu,
-        values = {"cpu": cpu},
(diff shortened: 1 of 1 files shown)


brentleyjones (dismissed):

keith:
@brentleyjones did your testing work

brentleyjones:
@keith I never got around to testing this yet, sorry. I did a search though and I don't think we are using these constraints, unless they are used implicitly somewhere.

keith:
i think you'd have to be using them directly, i don't think we use them internally

brentleyjones (approved):

PR #402, opened 2025-04-17
cc_binary: support .dSYM generation

benjivos:
Add support for generating a .dSYM file if this is requested using `--apple_generate_dsym`. This change contains the following patches:

* Add `DSYM_HINT` flags to be picked up by `wrapped_clang` to generate .dSYM bundles on request.
* In `wrapped_clang.cc` make sure the linked binary is stripped after generating the .dSYM bundle first to make the .dSYM bundle contain useful information.

Adopts [changes](https://github.com/bazelbuild/bazel/pull/25881) in bazel to support .dSYM files (as part of fixing https://github.com/bazelbuild/bazel/issues/16893).

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -868,7 +868,7 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
                 actions = _DYNAMIC_LINK_ACTIONS,
                 flag_groups = [
                     flag_group(
-                        flags = ["-Wl,-S"],
+                        flags = ["STRIP_DEBUG_SYMBOLS"],
                         expand_if_available = "strip_debug_symbols",
                     ),
                 ],
@@ -2062,13 +2062,15 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
                 flag_groups = [flag_group(flags = ["-g"])],
             ),
             flag_set(
-                actions = ["objc-executable"],
+                actions = _DYNAMIC_LINK_ACTIONS,
                 flag_groups = [
                     flag_group(
                         flags = [
-                            "DSYM_HINT_LINKED_BINARY=%{linked_binary}",
+                            "DSYM_HINT_LINKED_BINARY=%{output_execpath}",
                             "DSYM_HINT_DSYM_PATH=%{dsym_path}",
                         ],
+                        # We need to check this for backwards compatibility with bazel 7
+                        expand_if_available = "dsym_path",
                     ),
                 ],
             ),
--- a/crosstool/wrapped_clang.cc
+++ b/crosstool/wrapped_clang.cc
@@ -260,13 +260,15 @@ static std::unique_ptr<TempFile> WriteResponseFile(
 void ProcessArgument(const std::string arg, const std::string developer_dir,
                      const std::string sdk_root, const std::string cwd,
                      bool relative_ast_path, std::string &linked_binary,
-                     std::string &dsym_path, std::string toolchain_path,
+                     std::string &dsym_path, bool &strip_debug_symbols,
+                     std::string toolchain_path,
                      std::function<void(const std::string &)> consumer);
 
 bool ProcessResponseFile(const std::string arg, const std::string developer_dir,
                          const std::string sdk_root, const std::string cwd,
                          bool relative_ast_path, std::string &linked_binary,
-                         std::string &dsym_path, std::string toolchain_path,
+                         std::string &dsym_path, bool &strip_debug_symbols,
+                         std::string toolchain_path,
                          std::function<void(const std::string &)> consumer) {
   auto path = arg.substr(1);
   std::ifstream original_file(path);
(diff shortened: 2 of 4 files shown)


crosstool/wrapped_clang.cc
```diff
@@ -509,16 +519,31 @@ int main(int argc, char *argv[]) {
...
-                                            dsym_path,
-                                            "--flat",
-                                            "--no-swiftmodule-timestamp"};
+  // We should generate a .dSYM bundle even if a full path is passed to a flat
+  // .dSYM file for backwards compatibility
+  const std::string bundle = ".dSYM";
+  size_t pos = dsym_path.find(bundle);
+  dsym_path = pos == std::string::npos
+                  ? dsym_path
+                  : dsym_path.substr(0, pos + bundle.length());
```
keith, line 528: I don't think this logic is doing what you want, this command fails for me:

```
bazel build test/test_data:macos_binary --apple_generate_dsym
```

AFAIUI there are only 2 valid patterns for the original dsym_path, either:

```
bazel-out/darwin_arm64-dbg/bin/test/test_data/cc_test_binary.dSYM
```

or:

```
bazel-out/darwin_x86_64-fastbuild-macos-x86_64-min13.0-applebin_macos-ST-896069e7915a/bin/test/test_data/macos_binary_bin.dwarf
```

in both of these cases I don't think we need to change the dsym_path? maybe in the latter case we should just add `--flat` back?

crosstool/wrapped_clang.cc
```diff
@@ -359,6 +360,14 @@ void ProcessArgument(const std::string arg, const std::string developer_dir,
     return;
   }
 
+  // Don't strip the binary directly when stripping is requested, as this will
+  // remove all debug information that will be required to generate a .dSYM
+  // bundle
+  if (arg == "-Wl,-S") {
```
keith, line 366: i think we can improve this by using an env var instead, but we can do that separately

keith (commented):
Can you add a test similar to https://github.com/bazelbuild/apple_support/blob/af8d93a33586c80bf2d245922ce31fcd109fe645/test/linking_tests.bzl#L217-L226 for this? I think it will have to be disabled until your bazel change lands but at least that would be useful to verify locally too.

nit: please run [pre-commit](https://pre-commit.com/) to format the code with clang-format

I think the test failure here is related to this change, not 100% sure

crosstool/wrapped_clang.cc
benjivos: Good point, I wasn't aware of the `apple_binary` rules. Adding `--flat` back seems to be a better option indeed. This also works for [rules_swift](https://github.com/bazelbuild/rules_swift/blob/d296e13ab1832ac739c4da99f2dc1d69a36789b9/swift/toolchains/xcode_swift_toolchain.bzl#L591). This now works with your commands above.

crosstool/cc_toolchain_config.bzl
```diff
@@ -2069,6 +2069,18 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
                     ),
                 ],
             ),
+            flag_set(
+                actions = all_link_actions,
+                flag_groups = [
+                    flag_group(
+                        flags = [
+                            "DSYM_HINT_LINKED_BINARY=%{output_execpath}",
```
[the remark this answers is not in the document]
benjivos: Seems to work indeed!

keith:
i think this looks good but we need to wait for the bazel change to make sure this is the right path

benjivos:
@Keith should we wait until bazel 8.3 is released? (at least I could enable the test then)

keith:
/i think we can test this now, want to rebase

keith:
not right now but you can mess with https://github.com/bazelbuild/apple_support/blob/7009b77c98a67d3fea081c9db4dbcee8effc3b7e/.bazelci/presubmit.yml#L17 to exclude this single target manually for 7.x only

benjivos:
@Keith I was only able to manually add the new test target to the bazel 8+ tasks in `.bazelci/presubmit.yml`. I tried otherwise, but kept getting these failures when running with bazel 7:

```
ERROR: /Volumes/Projects/apple_support/test/BUILD:29:19: in dsym_test rule //test:linking_generate_cpp_dsym_test:
Traceback (most recent call last):
	File "/Volumes/Projects/apple_support/test/rules/action_command_line_test.bzl", line 80, column 40, in _action_command_line_test_impl
		concatenated_args = " ".join(action.argv) + " "
Error: Invalid toolchain configuration: Cannot find variable named 'dsym_path'.
ERROR: /Volumes/Projects/apple_support/test/BUILD:29:19: Analysis of target '//test:linking_generate_cpp_dsym_test' failed
```

keith:
hrm are we sure that doesn't actually break building cc_binarys with 7.x for real? that error makes it sound like if we were hitting the new codepath in the crosstool that it would fail because really you need `expand_if_available` for 2 attributes

benjivos:
> hrm are we sure that doesn't actually break building cc_binarys with 7.x for real? that error makes it sound like if we were hitting the new codepath in the crosstool that it would fail because really you need `expand_if_available` for 2 attributes

I managed to remove the manual tag, and only exclude it as a test target for 7.x. I did this by only using `expand_if_available` for the `dsym_path` attribute, which seems to work. Let me know what you think.

PR #411, opened 2025-05-20
Consume the pipes before waiting for pid

Reflejo:
In general one shouldn't call waitpid until the child closed its end of the pipe. This works fine with libtool invocations that do not produce more than `_PC_PIPE_BUF` bytes of output but deadlocks when the pipe fills up.


Reflejo:
@keith 👀

keith:
thanks, probably fixes https://github.com/bazelbuild/apple_support/issues/415, relevant: https://github.com/bazelbuild/apple_support/commit/bd83afb707dd5c62d68c52e7354428018b2c04ef

keith (approved):

PR #412, opened 2025-05-27
Upstream: Add an "empty SDK" variant.

luispadron:
This will be used as the `//conditions:default` choice for the `xcode_version.sdk` attribute, ensuring that non-configured builds/queries don't fail early if they just happen to graze one of the Xcode targets.

PiperOrigin-RevId: 726034982

And:

PiperOrigin-RevId: 725333279

### DIFF
--- a/xcode/private/BUILD
+++ b/xcode/private/BUILD
@@ -65,7 +65,7 @@ config_setting(
 )
 
 config_setting(
-    name = "is_macosx",
+    name = "is_macos",
     constraint_values = [
         "@platforms//os:macos",
     ],


xcode/private/BUILD
```diff
@@ -61,7 +65,7 @@ config_setting(
 )
 
 config_setting(
-    name = "is_macos",
+    name = "is_macosx",
```
aaronsky, line 68: can we put this back?

luispadron, line 68:
```suggestion
    name = "is_macos",
```

aaronsky (approved):

PR #416, opened 2025-07-01
Add support for BAZEL_CONLYOPTS, BAZEL_CXXOPTS, BAZEL_LINKOPTS

AustinSchuh:
When doing multi-platform builds, using --cxxopt and friends ends up modifying the toolchain for both OSX, and any target.  rules_cc provides environmental variables to modify the toolchain (which can be modified through --repo_env), which get around this issue.  Add support here for them too.


keith:
can you rebase? i think the argument that the default toolchain does this is enough for us to do it here too, but what's the use case for this vs passing `--copt` somewhere? i feel like most folks find that easier to throw in their `.bazelrc`

keith (approved):

keith:
i think this is fine, likely the real change will be once we can use rules based toolchains here we'll expect users to add features manually for this kinda thing instead

AustinSchuh:
Oh, 100%, rule based toolchains would be a lot better here.  One day...

PR #425, opened 2025-08-28
Remove apple_crosstool_top

keith:
This was removed from bazel @ HEAD


keith:
note that in this repo this was just for tests, so all that matters is ci passing. the fix in rules_apple might have to be different

mattrobmattrob (approved):

PR #428, opened 2025-09-02
Pass -object_path_lto <path> linker flag for LTO builds (#420)

ma-oli:
The following commit (in 1.23.0) was reverted recently (in 1.23.1):
* commit 61b44d12df0834ae26d6e089271e14c6c9a239c8
* revert 4c51f10063687af77eabbbdc035124d798ca3a3f

This patch is a new attempt at getting that feature merged, but without the LTO related issues that came up in the first patch. The difference is that we're adding that feature unconditionally now, as it won't hurt linking actions when LTO is disabled anyway. And enabling the bazel lto features changes the linking workflow in a way that isn't necessary on Apple platforms.

Initial commit log below.
---

This is needed for LTO builds, this is the path to which linker writes one big object file after performing link time optimisations, and then this object file would be used during final linking to create the App binary, without this file binary would contain invalid debug symbols, more info - https://github.com/bazelbuild/rules_swift/issues/1529 .

Closes - https://github.com/bazelbuild/rules_swift/issues/1529

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -1023,6 +1023,38 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
         ],
     )
 
+    # We need to pass -object_path_lto in order to get debug symbols when building with LTO.
+    # In a perfect world, we would only pass it when the thin_lto or full_lto features are enabled.
+    # However, when these features are enabled, bazel will follow the linking steps expected by
+    # the traditional llvm tools, which doesn't work on Apple platforms.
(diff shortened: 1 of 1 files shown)


keith (approved):
confirmed that this is a no-op unless LTO is actually produced in the build. seems harmless to me.

sanju-naik:
Thank you @ma-oli @keith for adding this back 🙏 . So the issue was happening because of adding these features - https://github.com/bazelbuild/apple_support/pull/420/files#diff-f87c478e9c0e6197378202814ad77531a720f7f1553c522123ba85e12de5e45aR2636, which weren't compatible or needed more things when building on non-Apple platforms?

PR #430, opened 2025-09-05
Enable no_warn_duplicate_libraries for Xcode 15+

keith:
This is added in the repo rule if supported, but if your host is linux
querying the system ld isn't valid. If you are using full Xcode we can
just check the version. If you're using the CLT only you'd have to
enable it manually.


aaronsky (approved):

keith:
never heard if this was tested but seems safe enough, lmk if you hit issues

PR #431, opened 2025-09-09
apple_genrule(): stop using some legacy features

EdSchouten:
- rule()'s output_to_genfiles stopped being meaningful when bazel-genfiles/ and bazel-bin/ consolidated.

- ctx.actions.run()'s input_manifests argument is a legacy argument that is documented as being ignored.


EdSchouten:
Aaah, I had to expand the output a bit to get a proper error message. Silly that it doesn't do that by default.

PR #458, opened 2025-09-24
Remove absolute AST path functionality

keith:
I don't think this is needed anymore


brentleyjones:
> I don't think this is needed anymore

What makes you think that? Want to have more evidence 😄.

keith:
If you are able to debug your project and don't flip this off then I believe previous issues were fixed. Def interested to hear if you do

PR #552, opened 2026-03-24
Add rules based toolchain

keith:
This sets up replacing our existing cc toolchain setup
with a new rules based toolchain. These rules are much more
ergonomic and should make maintaining this toolchain easier
over time. Specifically many things that were previously done
with random custom API specific to these toolchains is now
done with typical rules patterns such as `select()`ing on
different platforms, and reference targets rather than arbitrary
strings. This also allows us to remove a few layers of templating
that was previously quite hard to follow. Now there is a single
instance of the underlying toolchain configuration, which is
transitioned per platform for the `select()`s to work.

My goal with this toolchain was to have an entirely empty diff
between the representation of the old ones, and the new one.
Given some random implementation details the diff is not
literally 0, but it is very close and I have audited the
entire diff to try to make sure there are no actual
behavior differences.

Because of ordering concerns we treat everything as features
just like the old toolchain. This requires a bit of managing
when the feature is off by default, because it's location in
`enabled_features` is where it will actually end up in the args
but we can't enable it there. In this case we put in a stub feature
that depends on the user facing feature string. We could potentially
mess with the ordering in the future, but I wanted to stick to
today's working version as closely as possible.


keith:
all the relevant rules_cc changes have landed, but we need a new release

aaronsky (approved):

PR #566, opened 2026-03-29
Fix dotd validation with bazel 9.x

keith:
With 9.x bazel's default output_base on macOS changed to
`~/Library/Caches`. Previously we included `~/Library` in the system
include path allowlist that is consulted for dotd validation. That
meant any file in the output_base would be assumed to be allowed even if
it wasn't correctly defined in BUILD files.

`~/Library` is used for `~/Library/Toolchains` for custom Swift
toolchain installations that don't require `sudo` to install.
Technically you can also install the CLT there, but I imagine that would
break many other things. We now scope this better to exclude
`~/Library/Caches`


aaronsky (approved):

keith:
notably without this change any headers covered by this path aren't correctly handled by bazel and header changes don't correctly invalidate downstream dependencies, which is very bad!

PR #569, opened 2026-04-03
Add supports_dynamic_library_deps

keith:
This is paired with https://github.com/bazelbuild/rules_cc/pull/682
which makes binaries correctly link to shared libs if that is
requested.


keith:
only relevant if that other PR merges as is

keith:
need to add to the rules based toolchain PR if so

PR #570, opened 2026-04-10
feat: Do not add objc, lkmod, lkmodc++, etc build when building iOS kernel extensions

YungRaj:
Support https://github.com/bazelbuild/rules_apple/pull/2911 by not adding objective c, kmod, kmodc++ libraries when building iOS kernel extensions.

1. Objective C doesn't work inside the iOS/macOS kernel
2. kmod and kmodc++ libraries are macOS only. The libkmod.a and libkmodc++.a artifacts in the macOS SDK will not build for iOS kernel extensions

Realistically the solution for 2 is that we provide the rest of the symbols that are provided by `libkmod.a` and `libkmodc++.a` ourselves in the kext source code ourselves.

Example symbols

`_start`
`_stop`
`_OSKextGetCurrentIdentifier`
`_OSKextGetCurrentLoadTag`
`_OSKextGetCurrentVersionString`

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -861,6 +861,7 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
             flag_set(
                 actions = _DYNAMIC_LINK_ACTIONS,
                 flag_groups = [flag_group(flags = ["-fobjc-link-runtime"])],
+                with_features = [with_feature_set(not_features = ["kernel_extension"])],
             ),
             flag_set(
                 actions = _DYNAMIC_LINK_ACTIONS,
@@ -1734,7 +1735,12 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
 
     # Kernel extensions for Apple Silicon are arm64e.
     if (ctx.attr.cpu == "darwin_x86_64" or
-        ctx.attr.cpu == "darwin_arm64e"):
+        ctx.attr.cpu == "darwin_arm64e" or
+        ctx.attr.cpu.startswith("ios")):
+        kext_flags = ["-nostdlib", "-Xlinker", "-kext", "-lcc_kext"]
+        if not ctx.attr.cpu.startswith("ios"):
+            kext_flags.extend(["-lkmod", "-lkmodc++"])
+
         kernel_extension_feature = feature(
             name = "kernel_extension",
             flag_sets = [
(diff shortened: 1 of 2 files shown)


crosstool/cc_toolchain_config.bzl
```diff
@@ -1734,22 +1735,20 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
 
     # Kernel extensions for Apple Silicon are arm64e.
     if (ctx.attr.cpu == "darwin_x86_64" or
-        ctx.attr.cpu == "darwin_arm64e"):
+        ctx.attr.cpu == "darwin_arm64e" or
+        ctx.attr.cpu.startswith("ios")):
+        kext_flags = ["-nostdlib", "-Xlinker", "-kext"]
+        if not ctx.attr.cpu.startswith("ios"):
+            kext_flags.extend(["-lkmod", "-lkmodc++", "-lcc_kext"])
```
keith: looks like cc_kext does exist for iOS too, although maybe it gets autolinked? why exclude that one?

test/linking_tests.bzl
```diff
@@ -251,11 +251,11 @@ def linking_test_suite(name):
         tags = [name],
         expected_argv = [
             "-nostdlib",
+            "-Xlinker",
+            "-kext",
```
keith, line 255: plz add a test for the new iOS behavior too

crosstool/cc_toolchain_config.bzl
YungRaj: You're right I forgot to check for libcc_kext in the iOS SDK. Fixed

test/linking_tests.bzl
YungRaj: Done.

test/linking_tests.bzl
```diff
@@ -251,17 +251,37 @@ def linking_test_suite(name):
...
+        tags = [name],
+        expected_argv = [
+            "-nostdlib",
             "-Xlinker",
             "-kext",
+            "-lcc_kext",
         ],
         not_expected_argv = [
             "-lc++",
+            "-objc",
```
keith:
```suggestion
            "-ObjC",
```

keith (approved):

keith:
lint, cla, and one comment and i should be good to merge

test/linking_tests.bzl
YungRaj: Done

Issue #574, opened 2026-04-21
Compiler warning when compiling crosstool/libtool.cc: [-Wunused-result]


jpc-sf:
Full warning:

```
INFO: From Executing apple_genrule @@apple_support+//crosstool:exec_libtool.target_config [for tool]:
external/apple_support+/crosstool/libtool.cc:288:11: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
  288 |     hasher(file);
      |     ~~~~~~^~~~~~
1 warning generated.
```

This function has no side effects, so this line can safely be removed.

PR #575, opened 2026-04-21
Remove unused hasher function

jpc-sf:
This function is marked with [[nodiscard]], which generates a warning because the result is unused.

Issue: Compiler warning when compiling crosstool/libtool.cc: [-Wunused-result] #574 https://github.com/bazelbuild/apple_support/issues/574

### DIFF
--- a/crosstool/libtool.cc
+++ b/crosstool/libtool.cc
@@ -256,9 +256,6 @@ void createSymlinks(std::filesystem::path temp_directory,
   for (auto file : files) {
     std::filesystem::path path = file;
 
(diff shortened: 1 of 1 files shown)


keith (approved):
thanks! just to note you must be on an internal SDK or something, it doesn't seem like this is marked `[[nodiscard]]` yet.

jpc-sf:
my pleasure. yeah, it looks like some STL implementations add `[[nodiscard]]` to some functions, even though the standard declaration doesn't have it. afaict, this is permitted by the standard.

PR #583, opened 2026-05-15
Run `realpath` on the `$DEVELOPER_DIR` and `$SDKROOT` paths in the substitution test.

keith:
In some cases, we've observed inconsistencies among the paths; sometimes we get the symlink SDK (`MacOSX.sdk`) and sometimes we get the versioned directory (`MacOSX14.2.sdk`). Either is valid, but make the test deterministic by always running `realpath`.

PiperOrigin-RevId: 613179143
(cherry picked from commit 8edac7cb67cb07647178f7d0f1d1a51ed62abd8c)

### DIFF
--- a/test/apple_support_test.bzl
+++ b/test/apple_support_test.bzl
@@ -26,8 +26,8 @@ set -eu
 
 OUTPUT_FILE="$1" ; shift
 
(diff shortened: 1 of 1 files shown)


keith:
we've never seen these issues but harmless in these tests

adincebic (approved):

PR #606, opened 2026-06-17
Move ApplePlatformInfo here

keith:
This will be used to reduce the reliance on the apple fragment which
will eventually be removed.

(cherry picked from commit 4338378cedc266a6da52494f095bf77ec93defc7)


aaronsky (approved):

keith:
it doesn't break rules_apple as is https://github.com/bazelbuild/rules_apple/pull/3006 even though we do need to change rules_apple to use this one instead of its own

PR #611, opened 2026-06-17
Remove for_bazel_tests

keith:
Doesn't seem like we need these anymore. Downstream repos don't use them
either


keith:
https://github.com/bazelbuild/apple_support/commit/bbbacf466fbe33976434f76269fa964988254d9d

PR #612, opened 2026-06-17
Add gazelle for bzl_library generation

keith:


keith:
there's some random reordering diff here that i submitted a fix for, but minor

aaronsky (approved):

PR #614, opened 2026-07-02
Make sanitizer builds respect -c for -g*

### LINKED ISSUE #613: Builds with sanitizers are not debuggable
As soon as `--features=ubsan` or `--features=asan` is given, the resulting binary cannot be debugged as it's missing needed debug information, so evaluation of variables in LLDB fails, although stepping through code works.

With the [help of Gemini](https://gemini.google.com/share/d/1prn_OrtaxohmVK7h_zaxYv5Ckmua5sdB?usp=sharing), I tracked down the culprit to be the [`-gline-tables-only`](https://github.com/bazelbuild/apple_support/blob/main/toolchain/sanitizers/BUILD#L60) flag that is added whenever any sanitizer is enabled. This makes debugging impossible as the debugger cannot evaluate variable values.

Gemini suggested the following workaround: `--action_env=CCC_OVERRIDE_OPTIONS="x-gline-tables-only +-g +-fstandalone-debug"` which I can confirm that it works, although it's not very elegant. Another workaround would be to patch the `apple_support` and use the patched version internally, which is also cumbersome since we have multiple bazel-managed repositories and an internal bazel registry, so simple `single_version_override` tricks don't work for us.

Hopefully, the `-gline-tables-only` flag can be removed from the list of default flags appended when the sanitizer feature is enabled in the official version of `apple_support`.

keith:
Previously for `--features=asan` you got `-gline-tables-only`. This
isn't ideal if you actually want to debug the sanitized code. Now it
respects `--compilation_mode` so with `-c dbg` you get `-g` and with
`-c fastbuild` you get `-gline-tables-only`.

This is a tiny bit weird because we don't pass any `-g` for `fastbuild`
normally.

Fixes https://github.com/bazelbuild/apple_support/issues/613

### DIFF
--- a/crosstool/cc_toolchain_config.bzl
+++ b/crosstool/cc_toolchain_config.bzl
@@ -2153,18 +2153,11 @@ please file an issue at https://github.com/bazelbuild/apple_support/issues/new
                     ACTION_NAMES.objc_compile,
                     ACTION_NAMES.objcpp_compile,
                 ],
-                flag_groups = [
-                    flag_group(
-                        flags = [
-                            "-fno-omit-frame-pointer",
-                            "-fno-sanitize-recover=all",
-                        ],
-                    ),
-                ],
+                flag_groups = [flag_group(flags = ["-gline-tables-only"])],
                 with_features = [
-                    with_feature_set(features = ["asan"]),
-                    with_feature_set(features = ["tsan"]),
-                    with_feature_set(features = ["ubsan"]),
+                    with_feature_set(features = ["asan", "fastbuild"]),
+                    with_feature_set(features = ["tsan", "fastbuild"]),
+                    with_feature_set(features = ["ubsan", "fastbuild"]),
                 ],
             ),
             flag_set(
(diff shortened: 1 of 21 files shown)


DoDoENT:
A matter of fact was that today I actually wanted to debug `fastbuild` code - I had some issue that wasn't reproducing without optimizer.

keith:
i think at some point there are too many combos and you'll have to do `--features=asan --copt=-whatever`, so I guess this fix is more about the defaults trying to be as useful as possible

DoDoENT:
> i think at some point there are too many combos and you'll have to do `--features=asan --copt=-whatever`, so I guess this fix is more about the defaults trying to be as useful as possible

That won't work and was the reason I raised the issue in the first place.

I have `common --copt=-g` for all my builds, including production (so that symbols can be stripped and binary with symbols retained for future symbolification of crash traces), yet it didn't work.

As described [by Gemini](https://gemini.google.com/share/d/1prn_OrtaxohmVK7h_zaxYv5Ckmua5sdB?usp=sharing), the problem is that Bazel always puts flags that come from features *last*, thus overriding both `--copt` and `--per_file_copt`. The only workaround for clang is using undocumented environment variable `CCC_OVERRIDE_OPTIONS` which can then override options set by bazel. Without this, the clang invocation line looks something like this: 

```
-g -fsanitize=address -gline-tables-only
```

which then overrides the former `-g` option and makes debugging impossible.

keith:
ah i don't think the not being able to override the order was explicit in the original report. i think we should probably fix that separately from this: https://github.com/bazelbuild/apple_support/pull/615

DoDoENT:
> ah i don't think the not being able to override the order was explicit in the original report. i think we should probably fix that separately from this: #615

Heh 😅 , I linked the transcript of my chat with Gemini, who convinced me that flag order is a Bazel thing, not a toolchain thing.

I'm glad it's possible to allow user flags to override toolchain feature flags directly from the toolchain config.

adincebic (approved):

keith:
other PR needs to be merged too, but i think with these combined we should be good

PR #616, opened 2026-07-04
Support hermetic developer directories in the crosstool

AttilaTheFun:
Two small changes needed by [hermetic_apple_toolchains](https://github.com/AttilaTheFun/hermetic_apple_toolchains), which registers verbatim Xcode.app copies inside external repositories and drives the build with a path-valued `--xcode_version` (an absolute developer directory path, which Bazel uses directly as `DEVELOPER_DIR`).

- `configure_osx_toolchain` gains an `extra_include_dirs` parameter so a wrapping repository rule can allow builtin includes from the external repository root, where the vendored Xcodes live. (Without this, headers resolved from the hermetic SDKs are flagged as undeclared inclusions, since they are outside the default /Applications and /Library include roots.)
- `cc_toolchain_config.bzl` treats a path-valued Xcode version as the latest Xcode in its two inline `>=` feature checks (`no_warn_duplicate_libraries`, `reproducible_linker_flag`) instead of failing with an unparseable dotted version.

The path-as-version semantics in the second bullet are not new — this extends an existing, Apple-contributed convention to the last place that still hard-fails on it:

- Bazel itself: a path-valued `XCODE_VERSION_OVERRIDE` is used directly as `DEVELOPER_DIR` ("Directly use version as DEVELOPER_DIR when a path is passed", [`XcodeLocalEnvProvider.java`](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/exec/local/XcodeLocalEnvProvider.java)).
- apple_support: [`xcode_support.is_xcode_at_least_version`](https://github.com/bazelbuild/apple_support/blob/master/lib/xcode_support.bzl#L92) returns `True` for path-valued versions since bazelbuild/apple_support#481 ("We're gonna assume that if the user passes Xcode as a directory, they're ok managing the features themselves, and we'll assume they're on the latest and greatest Xcode").
- rules_swift: the toolchain's [`_is_xcode_at_least_version`](https://github.com/bazelbuild/rules_swift/blob/master/swift/toolchains/xcode_swift_toolchain.bzl#L722) does the same since bazelbuild/rules_swift#1597.

The new helper in `cc_toolchain_config.bzl` mirrors those implementations; the two inline comparisons were added later (for the Xcode 15 and Xcode 26 linker flags) and never picked up the convention. Without this, analysis fails with a Starlark error when comparing the path against `apple_common.dotted_version("15.0")`.

### DIFF
--- a/crosstool/osx_cc_configure.bzl
+++ b/crosstool/osx_cc_configure.bzl
@@ -147,11 +147,15 @@ def _succeeds(repository_ctx, *args):
 def _copy_file(repository_ctx, src, dest):
     repository_ctx.file(dest, content = repository_ctx.read(src))
 
-def configure_osx_toolchain(repository_ctx):
+def configure_osx_toolchain(repository_ctx, extra_include_dirs = []):
     """Configure C++ toolchain on macOS.
 
     Args:
       repository_ctx: The repository context.
+      extra_include_dirs: Additional directories to allow builtin includes
+          from, for toolchains and SDKs that live outside of the standard
+          Xcode and Command Line Tools installation directories (for example
+          hermetic toolchains vendored into external repositories).
 
     Returns:
       Whether or not configuration was successful
@@ -191,6 +195,7 @@ def configure_osx_toolchain(repository_ctx):
         features.append("reproducible_linker_flag")
 
     escaped_include_paths = _get_escaped_xcode_cxx_inc_directories(repository_ctx, xcode_toolchains)
+    escaped_include_paths += [escape_string(directory) for directory in extra_include_dirs]
     escaped_cxx_include_directories = []
     for path in escaped_include_paths:
         escaped_cxx_include_directories.append(("            \"%s\"," % path))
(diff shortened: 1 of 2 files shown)


crosstool/osx_cc_configure.bzl
```diff
@@ -147,11 +147,15 @@ def _succeeds(repository_ctx, *args):
 def _copy_file(repository_ctx, src, dest):
     repository_ctx.file(dest, content = repository_ctx.read(src))
 
-def configure_osx_toolchain(repository_ctx):
+def configure_osx_toolchain(repository_ctx, extra_include_dirs = []):
```
keith, line 150: this api isn't meant to be called externally and is being replaced by the rules based toolchain. can you use that one instead?

also i believe the value you're passing to this in your current rules is wrong and will ignore everything in external, which will break all header discovery https://github.com/bazelbuild/bazel/issues/29613

AttilaTheFun: oh let me see if this is even necessary with the rules based implementation. if not I can just drop it or else fix it another way

AttilaTheFun:
Closing in favor of the rules-based toolchain per the review — migrating to it removed the need for both changes here. The one remaining gap (include validation for vendored developer dirs) is addressed additively in the follow-up PR.

PR #627, opened 2026-07-21
Remove -objc_abi_version linker flag

keith:
This only affected non 32 bit macOS and doesn't seem to be passed by
Xcode anymore unless you set it to not 2, in which case ld rejects it
anyways.


adincebic (approved):
Wasn't expecting this many deletions. Nice.

PR #632, opened 2026-08-20
Enable module_map_without_extern_module by default

### LINKED ISSUE #631: C++ layering_check reports false missing-module errors on macOS through recursive extern module paths
### Summary

With Bazel 8.7.0, gRPC 1.83.0, and Protobuf 35.1, a minimal C++ consumer fails
under `layering_check` with messages claiming gRPC lacks modules exporting
generated UPB headers. The reported owner maps are already direct dependencies
and are explicitly present on the failing compile argv.

I reproduced this against Apple Support 2.8.0 and pinned current main commit
`8a5de128485e553bd35b8cd881b43480708f43ec`, selecting both the legacy and
rules-based toolchains:

| Apple Support | toolchain mode | result |
|---|---|---|
| 2.8.0 | legacy | fails |
| 2.8.0 | rules-based | fails |
| main at `8a5de128` | legacy | fails |
| main at `8a5de128` | rules-based | fails |

All four cells reach diagnostics for:

```text
envoy/type/v3/percent.upb.h
google/protobuf/duration.upb.h
google/protobuf/struct.upb.h
google/protobuf/wrappers.upb.h
```

### Minimal reproduction

`.bazelversion`:

```text
8.7.0
```

`.bazelrc`:

```text
common --enable_bzlmod
common --repo_env=APPLE_SUPPORT_LAYERING_CHECK_BETA=1

build --features=layering_check
build --cxxopt=-std=c++17
build --host_cxxopt=-std=c++17
```

`MODULE.bazel` for the 2.8.0 cells:

```starlark
module(name = "grpc_xds_layering_repro", version = "0.0.0")

bazel_dep(name = "apple_support", version = "2.8.0")
bazel_dep(name = "grpc", version = "1.83.0")
bazel_dep(name = "protobuf", version = "35.1")
```

For the current-main cells, append:

```starlark
git_override(
    module_name = "apple_support",
    commit = "8a5de128485e553bd35b8cd881b43480708f43ec",
    remote = "https://github.com/bazelbuild/apple_support.git",
)
```

Run both toolchain modes on macOS:

```sh
bazelisk build @grpc//src/core:grpc_xds_client \
  --repo_env=APPLE_SUPPORT_RULES_BASED_TOOLCHAIN=0 \
  --keep_going --verbose_failures

bazelisk build @grpc//src/core:grpc_xds_client \
  --repo_env=APPLE_SUPPORT_RULES_BASED_TOOLCHAIN=1 \
  --keep_going --verbose_failures
```

Verified host: macOS 26.5.2 arm64, Xcode 26.6, Apple clang 21.0.0
(`clang-2100.1.1.101`), macOS SDK 26.5, and `PATH_MAX=1024`.

### Root cause evidence

This matches LLVM issue #147220 and the fix merged in LLVM PR #193691. Older
Clang constructs nested relative `extern module` paths without normalizing
`..`; a sufficiently deep graph exceeds macOS `PATH_MAX`, fails to load an
existing map, and later reports a misleading `-fmodules-strict-decluse` error.

In this reproduction:

- the failing argv contains `-fmodules-strict-decluse` and each relevant owner
  `.cppmap`;
- gRPC's target map has direct `use` and `extern module` entries for those
  owners, and the owner maps textually export the diagnosed headers;
- 678 generated maps contain 6,936 extern edges;
- emulating the old loader from the 83-byte root reaches 5,922 bytes and first
  exceeds the 1,024-byte host limit at hop 12;
- removing recursive extern declarations makes the target pass while retaining
  strict decluse;
- a deliberately invalid direct include of a transitive-only header still
  fails, confirming that actual layering enforcement remains active.

### Possible compatibility bridge

Could C++ `layering_check` activate the existing rules_cc
`module_map_without_extern_module` generator mode in both Apple Support
toolchain implementations? Dependent maps are already passed explicitly with
`-fmodule-map-file`, so the recursive extern declarations are redundant in this
mode.

A two-line legacy-toolchain prototype passes all controls, but it is not a
complete current-main patch: the matrix above demonstrates that the rules-based
path must also be covered. The source split suggests that the reusable
rules-based feature may belong in rules_cc while Apple Support retains the
legacy feature and the dual-mode integration test. Maintainer guidance on that
boundary would be welcome.

LLVM's path normalization remains the canonical compiler fix; this proposal is
a compatibility bridge for released Apple Clang versions.

Related:

- https://github.com/llvm/llvm-project/issues/147220
- https://github.com/llvm/llvm-project/pull/193691
- https://github.com/open-telemetry/opentelemetry-cpp/pull/4164#issuecomment-2968452453

keith:
Fixes https://github.com/bazelbuild/apple_support/issues/631

Apple clang will eventually have https://github.com/llvm/llvm-project/pull/193691 which
should fix the original issue here, but as is not with Xcode 27.0 or below


keith:
seems like there is no downside

adincebic (approved):

PR #635, opened 2026-08-28
Let the rules-based cc_toolchain be driven by a non-Xcode clang

AustinSchuh:
The macro is already parameterized by tool_map and sysroot_feature, but targeting macOS with anything other than wrapped_clang does not work: everything behind select(//configs:apple) assumes the Xcode configuration is the source of truth.

Most of that needs no API. The __BAZEL_* placeholders and the STRIP_DEBUG_SYMBOLS / LINKED_BINARY= / DSYM_HINT_DSYM_PATH= sentinels are the toolchain's argument protocol, which a consumer's tool has to speak the way wrapped_clang does (now documented on tool_map), and layering_check works as-is since a plain clang ignores APPLE_SUPPORT_MODULEMAP.

What remains becomes a parameter, each defaulting to current behavior:

  - target_from_xcode = False uses the "target" argument instead of the Xcode-derived triple, for a toolchain that pins its own.
  - flags_from_env = False drops the BAZEL_*OPTS features, whose -std=c++17 default would override a toolchain's own standard and which cannot differ between two toolchains in one workspace.
  - coverage_instrumentation = False leaves the coverage-map-format features out for a consumer that instruments from its own args, since Bazel picks the gcc format by default.
  - apple_linker_flags = False drops -no_warn_duplicate_libraries and -reproducible, which ld64.lld only accepts from LLVM 19 on.
  - extra_{enabled,known}_features and extra_include_directories as attributes: the label_flags are global, so a repo rule stamping out more than one toolchain cannot vary them.
  - dynamic_runtime_lib / static_runtime_lib pass through to cc_toolchain, which is how a sanitizer runtime dylib reaches a test's runfiles.

Also make //toolchain:dynamic_toolchain_info public: the macro references it inside the Apple branch of a select(), so an Apple-targeting consumer fails analysis without it.

//test/... is unchanged.

### DIFF
--- a/toolchain/BUILD
+++ b/toolchain/BUILD
@@ -306,7 +306,10 @@ cc_args(
 
 dynamic_toolchain_info(
     name = "dynamic_toolchain_info",
-    visibility = ["//:__subpackages__"],
+    # The `cc_toolchain` macro references this from inside a select() on
+    # //configs:apple, so an external consumer of the macro that targets an
+    # Apple platform needs to be able to see it.
+    visibility = ["//visibility:public"],
 )
 
 xcode_execution_info(
@@ -333,6 +336,18 @@ negatable_feature(
         "//configs:apple": [":dynamic_toolchain_info"],
         "//conditions:default": [],
     }),
+)
+
+# As `default_required_flags`, but without the Xcode-derived `-target`, for a
+# toolchain that pins its own triple. Its consumer supplies `--target=` in its
+# own args, the way //test/linux_toolchain does.
+negatable_feature(
+    name = "plain_required_flags",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:link_actions",
+        "@rules_cc//cc/toolchains/actions:compile_actions",
+    ],
+    args = ["-no-canonical-prefixes"],
 )
 
 negatable_feature(
--- a/toolchain/BUILD
+++ b/toolchain/BUILD
@@ -183,12 +183,28 @@ cc_args_list(
     name = "default_compile_flags",
     args = [
         ":objcpp_default_flags",
+        ":no_cxx_modules_flags",
         "//toolchain/sanitizers:no_asan_compile_flags",
         ":always_compile_flags",
         ":fastbuild_compile_flags",
         ":opt_compile_flags",
         ":dbg_compile_flags",
     ],
+)
+
+# Bazel's layering_check is built on Clang module maps, which -std=c++20
+# breaks: from Clang 14 on the driver switches to C++20 modules semantics
+# unless told otherwise, and Bazel does not support C++20 modules yet.
+# https://github.com/llvm/llvm-project/commit/0556138624edf48621dd49a463dbe12e7101f17d
+cc_args(
+    name = "no_cxx_modules_flags",
+    actions = ["@rules_cc//cc/toolchains/actions:cpp_compile_actions"],
+    args = [
+        "-Xclang",
+        "-fno-cxx-modules",
+        "-Wno-module-import-in-extern-c",
+    ],
+    requires_any_of = ["@rules_cc//cc/toolchains/args/layering_check:use_module_maps"],
 )
 
 cc_args(
--- a/toolchain/BUILD
+++ b/toolchain/BUILD
@@ -338,18 +338,6 @@ negatable_feature(
     }),
 )
 
-# As `default_required_flags`, but without the Xcode-derived `-target`, for a
-# toolchain that pins its own triple. Its consumer supplies `--target=` in its
-# own args, the way //test/linux_toolchain does.
-negatable_feature(
-    name = "plain_required_flags",
-    actions = [
-        "@rules_cc//cc/toolchains/actions:link_actions",
-        "@rules_cc//cc/toolchains/actions:compile_actions",
-    ],
-    args = ["-no-canonical-prefixes"],
-)
-
 negatable_feature(
     name = "__apply_simulator_compiler_flags",
     actions = [
--- a/toolchain/cc_toolchain.bzl
+++ b/toolchain/cc_toolchain.bzl
@@ -15,7 +15,6 @@ def cc_toolchain(
         supports_header_parsing,
         tool_map,
         flags_from_env = True,
-        coverage_instrumentation = True,
         llvm_version = None,
         extra_enabled_features = None,
         extra_known_features = None,
@@ -44,12 +43,6 @@ def cc_toolchain(
             toolchain that sets its own C++ standard has to turn this off --
             the environment is read once per build, so it also cannot differ
             between two toolchains in the same workspace.
-        coverage_instrumentation: Whether to contribute the gcc and llvm
-            coverage-map-format instrumentation flags, letting Bazel pick the
-            format (the gcc one unless --experimental_use_llvm_covmap).
-            Defaults to True. Set it to False for a toolchain that instruments
-            from its own args -- e.g. one whose coverage tooling is llvm-only
-            and so always wants the llvm format.
         llvm_version: The LLVM version of the tools in `tool_map`, e.g.
             "17.0.6", when they come from a plain LLVM distribution rather
             than from Xcode. Link flags that a distribution's `ld64.lld` only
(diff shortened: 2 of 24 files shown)


toolchain/cc_toolchain.bzl
```diff
@@ -22,14 +31,63 @@ def cc_toolchain(
...
+            `ld64` accepts -- `-no_warn_duplicate_libraries` and
+            `-reproducible`. Defaults to True, which is right for a toolchain
+            that links with the `ld` from Xcode. `ld64.lld` rejects an argument
+            it does not know, so a toolchain linking with a bundled `ld64.lld`
+            has to turn this off for the LLVM versions that predate it.
+        coverage_instrumentation: Whether to contribute the gcc and llvm
+            coverage-map-format instrumentation flags. Defaults to True. Set it
+            to False to instrument from the consumer's own args instead --
+            which format Bazel picks is otherwise up to Bazel, and the gcc one
+            is the default.
```
keith: what are you passing instead here? last i looked I was thinking we did the pretty vanilla bazel thing here

toolchain/cc_toolchain.bzl
```diff
@@ -22,14 +31,63 @@ def cc_toolchain(
...
+            placeholders, and the sentinel args `STRIP_DEBUG_SYMBOLS`,
+            `LINKED_BINARY=...` and `DSYM_HINT_DSYM_PATH=...` that stand for
+            work the tool does around the compiler. A `tool_map` that does not
+            drive `wrapped_clang` itself has to supply a tool that resolves the
+            placeholders and acts on the sentinels the same way.
+        target_from_xcode: Whether to take the target triple from the Xcode
+            configuration on Apple platforms, which is what this repo's own
+            toolchain does and which ignores `target`. Set it to False to use
+            `target` on every platform, as a toolchain that pins its own triple
+            (and passes its own deployment-target flag) needs.
```
keith: what are you trying to pass here vs what would the default be if this wasn't set? I believe our logic here is just reading the values coming from bazel really so I'd hope that would be ~the same

toolchain/BUILD
```diff
@@ -335,6 +338,18 @@ negatable_feature(
     }),
 )
 
+# As `default_required_flags`, but without the Xcode-derived `-target`, for a
+# toolchain that pins its own triple. Its consumer supplies `--target=` in its
+# own args, the way //test/linux_toolchain does.
+negatable_feature(
```
keith: I'm hoping we can figure out how to remove this and use the default `-target` thing provided above, but if we have to keep this can we just add it always, remove `-no-canonical-prefixes` from the one above, and name this feature `__no_canonical_prefixes`, since it should likely never be disabled

toolchain/cc_toolchain.bzl
```diff
@@ -22,14 +31,63 @@ def cc_toolchain(
...
+            True. Note that `BAZEL_CXXOPTS` defaults to `-std=c++17`, so a
+            toolchain that sets its own C++ standard has to turn this off --
+            the environment is read once per build, so it also cannot differ
+            between two toolchains in the same workspace.
+        apple_linker_flags: Whether to pass the link flags that only Apple's
+            `ld64` accepts -- `-no_warn_duplicate_libraries` and
+            `-reproducible`. Defaults to True, which is right for a toolchain
+            that links with the `ld` from Xcode. `ld64.lld` rejects an argument
+            it does not know, so a toolchain linking with a bundled `ld64.lld`
+            has to turn this off for the LLVM versions that predate it.
```
keith: hrm i don't love this one, the contract isn't very clear here since the values are supported upstream just maybe out of sync with the provided version. maybe we should try to do this based on llvm version instead? since then we could check if it was > a version and add it?

toolchain/cc_toolchain.bzl
```diff
@@ -22,14 +31,63 @@ def cc_toolchain(
...
+        coverage_instrumentation: Whether to contribute the gcc and llvm
+            coverage-map-format instrumentation flags. Defaults to True. Set it
+            to False to instrument from the consumer's own args instead --
+            which format Bazel picks is otherwise up to Bazel, and the gcc one
+            is the default.
+        extra_enabled_features: A `cc_feature_set` of extra features to enable,
+            in addition to the `//toolchain:extra_enabled_features` flag. A
+            toolchain that is generated more than once per workspace needs this
+            as an attribute, since the flag is global and cannot differ between
+            two toolchains.
```
keith:
```suggestion
        extra_enabled_features: A `cc_feature_set` of extra features to enable.
```

toolchain/cc_toolchain.bzl
```diff
@@ -22,14 +31,63 @@ def cc_toolchain(
...
+        extra_known_features: A `cc_feature_set` of extra features to make
+            known, in addition to the `//toolchain:extra_known_features` flag.
```
keith:
```suggestion
        extra_known_features: A `cc_feature_set` of extra features to make known.
```

toolchain/cc_toolchain.bzl
```diff
@@ -22,14 +31,63 @@ def cc_toolchain(
...
+        extra_include_directories: A `cc_args` of extra include directories, in
+            addition to the `//toolchain:extra_include_directories` flag.
```
keith:
```suggestion
        extra_include_directories: A `cc_args` of extra include directories.
```

toolchain/cc_toolchain.bzl
```diff
@@ -22,14 +31,63 @@ def cc_toolchain(
...
+        dynamic_runtime_lib: Passed through to `cc_toolchain`. The dynamic
+            library to link when `static_link_cpp_runtimes` is enabled, which
+            is how a sanitizer runtime dylib reaches a test's runfiles.
```
keith:
```suggestion
        dynamic_runtime_lib: Passed through to `cc_toolchain`. The dynamic
            library to link when `static_link_cpp_runtimes` is enabled.
```

toolchain/cc_toolchain.bzl
```diff
@@ -22,14 +31,63 @@ def cc_toolchain(
...
+    extra_enabled_features = [extra_enabled_features] if extra_enabled_features else []
+    extra_known_features = [extra_known_features] if extra_known_features else []
+    extra_include_directories = [extra_include_directories] if extra_include_directories else []
     _cc_toolchain(
         name = name,
         args = [
+            # An allowlist of the directories Xcode and the command line tools
+            # install into. It only widens what Bazel will accept as a builtin
+            # include, so it cannot break a build that does not need it, and
+            # any toolchain compiling against an Xcode SDK does need it.
```
keith:
```suggestion
```

toolchain/cc_toolchain.bzl
AustinSchuh: I like that better, done.

toolchain/BUILD
AustinSchuh: With some more prodding, the answer is yes.  Removed now.

toolchain/cc_toolchain.bzl
AustinSchuh: https://github.com/AustinSchuh/toolchains_llvm/tree/rules-based-darwin-toolchain is the prototype if you want to take a peek.

Correct me if I've got this wrong, but my understanding is that toolchains_llvm uses llvm coverage formats, and this uses the gcc format.  Adding this flag lets toolchains_llvm support only llvm and provide the features needed to do so.

toolchain/cc_toolchain.bzl
AustinSchuh: Thanks for the prod.  I started with making things match more identically and have been challenging each piece one by one.  After poking, looks like the answer for this one is "nope, not needed", apple_support's version is the same.

toolchain/cc_toolchain.bzl
keith: bazel's intent is that the gcc format is default, and then you pass the random flags to set it to llvm's format. but both are supported by clang as well

AustinSchuh: Makes sense.  I was able to drop this and get gcc based coverage working in toolchains_llvm.

AustinSchuh:
OK, everything should now be good to go.  I pulled in a couple of the flags from toolchains_llvm up here where they seemed better.  Happy to push those back into toolchains_llvm if you would prefer.

Thanks Keith!

keith (approved):

PR #637, opened 2026-09-14
Require macOS execution for Apple crosstool helpers

### LINKED ISSUE #636: force_exec selects Linux for Xcode compiler wrappers when Linux is the first execution platform
## Problem

With Linux registered before macOS as an execution platform, an iOS build fails during analysis of `@apple_support//crosstool:exec_wrapped_clang.target_config` (or `exec_wrapped_clang_pp.target_config`). The Apple C++ toolchain correctly selects macOS, but the intermediate `force_exec` rule independently selects Linux and transitions the compiler-wrapper dependency into a Linux target configuration.

The intended setup is a mixed Linux/macOS execution graph:

- Xcode-dependent compilation, linking, SDK discovery, and simulator operations execute on macOS.
- Portable tools and build actions can execute on Linux.
- A hermetic LLVM toolchain can cross-compile macOS executables such as `protoc` on Linux, and those executables can subsequently run on macOS while building the iOS app.

Making macOS the only execution platform, or disabling Linux for the entire iOS invocation, prevents the last use case. Execution constraints need to be attached to the Xcode-dependent helpers individually.

## Environment and reproduction

Reproduced with:

- Bazel 9.2.0
- apple_support 2.8.2, with `--repo_env=APPLE_SUPPORT_RULES_BASED_TOOLCHAIN=1`
- rules_apple 5.0.0 and rules_swift 4.0.1
- Hermetic LLVM module `llvm` 0.8.19
- Apple Silicon macOS host; Xcode 26.6.0 (17F113)
- Linux x86_64 and macOS arm64 execution platforms, in that order

Repository: [DoDoENT/bazel-playground](https://github.com/DoDoENT/bazel-playground/tree/7d97b198fffc61b9d32cee76a1bc0acc0faa2d28).

Using the repository's configured dependencies and execution backend, explicitly override its iOS-specific platform-list workaround:

```sh
bazel test //:ios --config=ios_simulator \
  --extra_execution_platforms=//platforms/linux:haswell_x86_64,//platforms/macos:arm64
```

The relevant platform constraints are `@platforms//os:linux` / `@platforms//cpu:x86_64` and `@platforms//os:macos` / `@platforms//cpu:arm64`. The first failure is during analysis, before any compiler-wrapper action executes.

## Actual behavior

```text
in apple_genrule rule @@apple_support+//crosstool:exec_wrapped_clang_pp.target_config:
  apple_platform_info = apple_support.platform_info_from_rule_ctx(ctx)
ERROR: A valid Apple platform constraint could not be found for target //crosstool:exec_wrapped_clang_pp.target_config

Check that you are building this target for a supported Apple platform (ios, macos, tvos, visionos, watchos) and that it is not accidentally being built with a default config, such as Linux.
```

Toolchain-resolution diagnostics showed the iOS C++ toolchain selecting `//platforms/macos:arm64`, followed by the unconstrained helper selecting `//platforms/linux:haswell_x86_64`.

## Cause and suggested patch

[`force_exec`](https://github.com/bazelbuild/apple_support/blob/main/crosstool/universal_exec_tool.bzl) has a `target` attribute with `cfg = "exec"`, but its own rule definition has no execution constraint or toolchain requirement that excludes Linux. The selected execution platform therefore becomes the target platform of the `.target_config` dependency.

The dependency is an `apple_genrule` that already requires macOS execution. That requirement does not propagate backwards to `force_exec` and does not repair the Linux **target** configuration that `force_exec` passed to it. The genrule queries Apple target-platform information during analysis, so it fails before its correctly constrained action could run.

A minimal guideline patch is:

```diff
--- a/crosstool/universal_exec_tool.bzl
+++ b/crosstool/universal_exec_tool.bzl
@@ -20,6 +20,7 @@
         ),
     },
     implementation = _force_exec_impl,
+    exec_compatible_with = ["@platforms//os:macos"],
 )
 
 def universal_exec_tool(*, name, out, srcs):
```

This constrains the Apple-specific helper rather than all targets whose output platform is macOS. Ordinary LLVM cross-compilation remains eligible to use Linux. The current upstream `force_exec` definition was also checked and still lacks this constraint.

If keeping `force_exec` generic is preferred, the equivalent constraint could be supplied at its Apple-only call sites; those uses should be audited together, including the compiler, libtool, and other crosstool wrappers.

## Validation and regression-test suggestion

Applied the one-line change in a temporary local module override, leaving Linux first. The original analysis failure disappeared and the iOS targets completed analysis. This exposed separate missing constraints in rules_apple's `environment_plist` and iOS runner rules.

After addressing those additional rules_apple constraints and a downstream `no-remote-exec` tag that had incorrectly forced a Linux-configured bundle tool to run locally on macOS, all five iOS simulator tests actually executed and passed. This is an integration result for the combined fixes, not a claim that this one patch alone fixes every mixed-platform issue.

A focused regression test should register Linux before macOS and verify that `force_exec` configures the Xcode compiler-wrapper dependencies for macOS. Ideally also retain a portable macOS tool built using a Linux-hosted cross-toolchain, to catch accidental blanket macOS execution restrictions.

DoDoENT:
When Linux is registered before macOS, `force_exec` selects Linux and configures its Xcode compiler-wrapper dependency for a Linux target platform. The dependency then fails during analysis when it queries Apple platform information, even though the surrounding iOS C++ toolchain correctly selected macOS.

Require macOS execution on `force_exec` so its exec transition configures the Apple crosstool helpers for macOS. This preserves Linux execution for ordinary macOS tools cross-compiled with hermetic LLVM.

Fixes #636.

Validation:
- `buildifier -mode=check -lint=warn crosstool/universal_exec_tool.bzl` and `git diff --check` passed.
- All five downstream iOS simulator tests executed and passed with Linux first and `--nocache_test_results`, using this exact checkout and the companion rules_apple constraints described in bazelbuild/rules_apple#3064.
- The downstream `no-remote-exec` tag was temporarily removed from the test macro because it also propagated to Linux-configured bundle actions. Simulator tests remained local and sequential through `exclusive`.

The integration result covers the combined Apple helper fixes; this PR addresses the earlier compiler-wrapper analysis failure.


DoDoENT:
Note: the test failure predates this patch.

keith:
please rebase after https://github.com/bazelbuild/apple_support/pull/638 lands

